PR #8, opened 2014-07-17
Fixing links to css and java script files for OC in reverse proxy enviro...

toskrip:
Implement fixes for scenario when running OC in reverse proxy environment.

In view data entry form was not able to display CRF header (by default it is hidden) because some java script and css files are not linked properly. This is fixed now.


toskrip:
I think you can close this pull request now. It seems to be corrected within OC-5916.

PR #49, opened 2014-09-16
Saml

davidbernick:


angood:
Hello David, 

After looking into the code we believe the changes break all other authentication/authorization models and only implements SAML. 

Is this something that you would be willing to address / update in your code (or perhaps you have)? 

Thanks, 
Alicia

PR #78, opened 2014-11-07
Fixing not updating Japanese choices on select box

mokjpn:
This commit fixes a problem discussed in
https://www.openclinica.com/forums#/discussion/15461

In summary, when OpenClinica runs on non-english locales on server and
users prefer English on their browser, choices of select box in “Create
Study” or “Update Study Details” are displayed in non-English language.
This is caused by lack of code which will update HashMaps of
internationalized resources.


angood:
hi, 

Thanks for submitting this. We'll review and get back to you!

Best, 
Alicia

PR #92, opened 2015-01-08
OC-5769: A study designer can view the status of PP study access request

sdibona:
A study designer can view the status of PP study access request

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/control/managestudy/UpdateStudyServletNew.java
+++ b/web/src/main/java/org/akaza/openclinica/control/managestudy/UpdateStudyServletNew.java
@@ -7,6 +7,9 @@
  */
 package org.akaza.openclinica.control.managestudy;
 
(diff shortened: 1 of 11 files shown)


web/src/main/java/org/akaza/openclinica/control/managestudy/UpdateStudyServletNew.java
```diff
@@ -649,4 +660,38 @@ private void updateInterviewerForSites(StudyBean studyBean, List<StudyBean> site
         }
 
     }
+    private String getPManageRegStatus(String studyOid) throws Exception
+    {
+		String pManageUrl = CoreResources.getField("portalURL");
```
kkrumlian: Looks like this code is duplicated in 2 different methods in the future please create one common method and reuse code. Also let's try to use RestTemplate going forward.

PR #93, opened 2015-01-09
Add debugging statements and enable PostgreSQL 9.x support

mjevans:
Modify the migration plans to avoid things that are un-necessary and which break PostgreSQL 9.x.

Add in some debug statements to make it more clear where configuration files are being loaded from as well as some SQL events when the log level is set to DEBUG.

I have tested these changes with PostgreSQL 9.4, testing against 8.x is not advised as all versions earlier than 9.x are EOL ( http://www.postgresql.org/support/versioning/ ).


angood:
hi, 

Thanks for submitting this. We'll review and get back to you!

Best, 
Alicia

PR #95, opened 2015-01-12
OC-6051  Hide system-created Participant (Data entry) user accounts 

jkeremian:


kkrumlian:
:+1: Looks good

core/src/main/resources/properties/useraccount_dao.xml
```diff
@@ -6,11 +6,11 @@
 	</query>
 	<query>
 		<name>findAll</name>
-		<sql>SELECT * FROM USER_ACCOUNT</sql>
+		<sql>SELECT * FROM USER_ACCOUNT WHERE USER_NAME NOT LIKE '%.SS_%'</sql>
```
kkrumlian: My advice here would be to create a findAllWithoutStudySubject. That way it is clear what you are returning.

core/src/main/resources/properties/useraccount_dao.xml
```diff
@@ -6,11 +6,11 @@
...
 	</query>
 	<query>
 		<name>findAllByLimit</name>
```
kkrumlian: My advice here would be to create a findAllByLimitWithoutStudySubject. That way it is clear what you are returning. No action required for now. In the future if we are working in this area let's change it

PR #96, opened 2015-01-13
Merge pull request #1 from OpenClinica/master

jacobrousseau:
Syncing my fork with OC-master


jacobrousseau:
Hallo Krikor, here is a pull request which we had to make due to a requirement we got from a security/privacy officer of one of our University Medical Centers. We must register more information on the exports (the person performing the export, the items exported and the subject's information exported). Could you have a look and assess if it can be included in the upcoming release? In the mean time we're going to add these 2 classes manually to the WAR

PR #98, opened 2015-01-13
OC-6091: OC can capture data from Open Rosa-compliant clients.

sdibona:
Added fix so /formList API will work with ODK Collect


kkrumlian:
:+1: Looks good

PR #106, opened 2015-01-23
OC-6201 PForm: Data not saved correctly if Single-select Pull Down item with long string and/or contains comma 

jkeremian:
checking in fix code for this reported bug
I'm removing the white spaces for item labels and values


kkrumlian:
:+1: Looks good

PR #107, opened 2015-01-23
OC-6202 Removing the implementation of creating new user accounts during PForm Data Submission

jkeremian:
Commented out the methods and their call that creates a new user account
in case if it does not exist during PForm data submission

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/service/PformSubmissionService.java
+++ b/core/src/main/java/org/akaza/openclinica/service/PformSubmissionService.java
@@ -154,7 +154,7 @@ public class PformSubmissionService {
 	 * @return
 	 * @throws Exception
 	 */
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/akaza/openclinica/service/PformSubmissionService.java
```diff
@@ -154,7 +154,7 @@ public String getInputUsername(StudyBean studyBean, StudySubjectBean studySubjec
 	 * @return
 	 * @throws Exception
 	 */
-	private UserAccountBean createUserAccount(UserAccountBean userAccountBean, StudyBean studyBean, StudySubjectBean studySubjectBean)
+/*	private UserAccountBean createUserAccount(UserAccountBean userAccountBean, StudyBean studyBean, StudySubjectBean studySubjectBean)
```
kkrumlian: In the future please delete rather than commenting out

PR #108, opened 2015-01-23
OC-6132a Registering a participant stores the participants info in OC

jkeremian:
adding a dummy password when creating participant user


kkrumlian:
:+1: Looks good

PR #109, opened 2015-01-23
OC-5802 Add column in Event Definition CRF (don't break CRUD) 

jkeremian:
I have added a new column in event_def_crf table for participant crf
I have updated the Bean and DAO files for CRUD operations


kkrumlian:
:+1: Looks good

PR #110, opened 2015-01-28
OC-6216  OCForm: Integer field validation fails

jkeremian:
I have converted the datatype from 'integer' into 'int' to resolve this
issue as Martin suggested that might be the cause.
I also added temporarily logger.info for the whole body , 
1. for rendering
2. saving


kkrumlian:
:+1:  Looks good

PR #117, opened 2015-02-23
OC-6203  Add a link in OpenClinica to access participate

jkeremian:


web/src/main/java/org/akaza/openclinica/control/MainMenuServlet.java
```diff
@@ -292,11 +293,22 @@ private void setupListStudySubjectTable() {
...
         factory.setEventDefintionCRFDAO(getEventDefinitionCRFDAO());
         factory.setStudyGroupDAO(getStudyGroupDAO());
+        factory.setStudyParameterValueDAO(getStudyParameterValueDAO());
         String findSubjectsHtml = factory.createTable(request, response).render();
         request.setAttribute("findSubjectsHtml", findSubjectsHtml);
     }
 
-    public StudyEventDefinitionDAO getStudyEventDefinitionDao() {
+    
+    public StudyParameterValueDAO getStudyParameterValueDAO() {
```
kkrumlian: Should this code eventually just live in SecureController seems like we are using it a lot. Perhaps in the future we should also put it in a filter.

kkrumlian:
:+1: Looks good

PR #120, opened 2015-03-02
OC-6273: Issues with authentication to OCUI

jkeremian:
Issues with authentication to OCUI
1. Site level users (CRC and Data Entry) are being rejected by OCUI
2. If an authorized user (CRC or Data Entry (Site and Study) is removed then he can still log in to OCUI


kkrumlian:
:+1: Looks good

PR #128, opened 2015-03-24
OC-6233: Study designer request access - Update UI.  Modified to call re...

sdibona:
...named REST service.


kkrumlian:
:+1: Looks good

PR #130, opened 2015-03-25
OC-6233: Study designer request access - Update UI.  Fixed bug with tabl...

sdibona:
...e borders not appearing for empty cells.  Specific to Firefox.


kkrumlian:
:+1: Looks good

PR #133, opened 2015-03-31
WS import bug, where random version of a response option is selected and causes error

mh11:
Wrong selection happens because of unsorted (random) order of versions returned from DB
and first one is picked.
This fix selects the version specific option.

### DIFF
--- a/ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
+++ b/ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
@@ -365,8 +365,8 @@ public class ImportCRFDataService {
                                     DisplayItemBean displayItemBean = new DisplayItemBean();
                                     displayItemBean.setItem(itemBean);
 
-                                    ArrayList<ItemFormMetadataBean> metadataBeans = itemFormMetadataDAO.findAllByItemId(itemBean.getId());
-                                    logger.debug("      found metadata item beans: " + metadataBeans.size());
+                                    ItemFormMetadataBean metadataVersionBean = itemFormMetadataDAO.findByItemIdAndCRFVersionId(itemBean.getId(), crfVersion.getId());
+                                    logger.debug(String.format("      found metadata version specific item beans: %s", metadataVersionBean));
                                     // groupOrdinal = the ordinal in item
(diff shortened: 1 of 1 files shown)


sdibona:
Hello.

I am reviewing the code changes and would like to understand the original error better.  Would you please describe in more details the steps to recreate the issue (or reference a JIRA ticket if one exists)?  For the item in question, what was different in the metadata between versions that eventually led to the error?  Can you provide a stack trace of the error that occurred?

Thank you,
Steve

mh11:
CONTEXT:
Assuming you have a CRF called 'MyCRF' and contains a single-select item with the name 'MY_ITEM' and two versions (1 and 2).
in Version 1, 'MY_ITEM' has a response set 'a,b,c' with the label "A_B_C".
In Version 2, 'MY_ITEM' has a response set 'x,y,z' with the label "X_Y_Z".

ISSUE:
In your current code, the call below retrieves all response sets (query listed further below)
 ->  itemFormMetadataDAO.findAllByItemId(itemBean.getId()) 

... and later chooses the first item in the list returned by the query (random order)
  ->  'metadataBeans.get(0);'

The Metadata selection is not version specific.

CONSEQUENCES:
When importing data using OC WS, I got RANDOM errors messages, that the submitted 'value' is not in the correct response set. (even though the correct Form version etc. was selected) 
That was annoying ... and how do you report such a random bug, which looks like to be a user issue?

FIX:
By requesting the Item Metadata by item ID and Version, the problem is solved:
-> itemFormMetadataDAO.findByItemIdAndCRFVersionId(...)

Hope this helps

-- SQL QUERY in OpenClinica used to retrieve response sets for all versions.
SELECT distinct m.*, rs.response_type_id, rs.label, rs.options_text,rs.options_values, cv.name as cvname, ig.name as group_label, igm.repeat_max, sec.title as section_name
FROM item_form_metadata m, response_set rs, crf_version cv, item_group_metadata igm, item_group ig, section sec
WHERE m.item_id= <ITEM_ID>
AND m.response_set_id = rs.response_set_id
AND cv.crf_version_id = m.crf_version_id
AND igm.item_id = m.item_id
AND ig.item_group_id = igm.item_group_id
AND sec.section_id = m.section_id

PR #140, opened 2015-04-03
OC-6176: PManage: Removed Event still available to Participant and entered data will be saved.  Initial checkin.

sdibona:


kkrumlian:
:+1: Looks good

PR #144, opened 2015-04-06
OC-6233: Study designer request access - Update UI.  Fixed issued when r...

sdibona:
...equesting hostnames that were already in use.


kkrumlian:
:+1: Looks good

PR #146, opened 2015-04-07
OC-6357: Copy Fixes for Request Access form.  Initial checkin.  Also rem...

sdibona:
...oved portalURL from datainfo.properties.


kkrumlian:
:+1: Looks good

PR #160, opened 2015-04-14
OC-showhideBug

kkrumlian:
Fix for Oops error , also added a statement not to show fields with
empty records


kkrumlian:
:+1: Looks good

PR #172, opened 2015-05-08
Addition of a footer to the notification mail sent after jobs complete

jacobrousseau:
Hallo Krikor,
this is an additional requirement from a security officer; a custom notification must be added in the email sent out to notify users of export jobs. It is combined with the previous pull request,

regards, Jacob


sdibona:
Hello Jacob,

I've reviewed the code changes and overall they look good.  My main concern is with putting a custom notification into datainfo.properties.  That file is generally reserved for major system/feature configuration settings.  

The footer message you want to change already exists as a configurable field in the page_messages.properties file and can be customized as needed.  Those property files are where custom messaging is designed to reside.  Is there a reason that wouldn't suffice for your security officer?

Thank you,
Steve

jacobrousseau:
Hi Steve,

the reason I choose the datainfo.properties is that it is 'accessible' without unjarring / unzipping the WAR. E.g. it is more accessible for system-administrators. Also I noticed that Krikor started pulling out the datainfo.properties file into a openclinica-config directory. 
I.e. a system-administrator without Java-knowledge can access this file. 
N.B. I'm would like to retract this pull-request and submit a new one. This will contain 3 functional changes:
- the change of the footer in the mail notifying that the export has changed (similar to the retracted pull-request)
- additional changes to also log the access to export files in the AccessFileServlet
- fix for OC-6466  Null Pointer Exception in RuleDesigner

PR #175, opened 2015-05-12
OC-6421: Set CRF status after import via SOAP WS.  Initial checkin.

sdibona:


kkrumlian:
Steve is there a reason why the files appear as if they've been rewritten ?

sdibona:
It's the Eclipse formatting file we're using.  It's changing all the whitespace.  You can view the github diff minus the whitespace by appending "?w=1" to the end of the url.

PR #179, opened 2015-05-14
OC-6412 Deleting event CRF will have status checks / updates Event

jkeremian:
status
Updated the code to address AC2 regarding to signed events


kkrumlian:
:+1: Looks good

PR #180, opened 2015-05-14
OC-6412  OC-6412 Deleting event CRF will have status checks / updates

jkeremian:
Event 
Updated the User for Event audit log record


kkrumlian:
:+1: Looks good

PR #186, opened 2015-05-21
OC-5994: Unable to asign user role due to data base column definition…

sdibona:
… for user_name has different lengths.  Initial checkin.


kkrumlian:
:+1: Looks good

PR #192, opened 2015-05-26
OC-5754 Rule Designer: For EventAction, If there is more than one Eve…

rooney:
…ntAction added to a target, RD crashes.

It's just missing an enum ActionType.EVENT on the OC-side, so I just added that one-liner fix


rooney:
Laura said on the ticket that my fix is not enough, so closing this and back to In Progress

PR #194, opened 2015-05-28
Update global_functions_javascript.js

iDrakos:
(Please also see pull request #195)

Function showSummaryBox(divObject,parentLinkObj,showText,hideText) is called though OC’s code passing divObject as an actual element or the ID of an element.

The old function code had two different variable declarations to handle the two different calling scenarios:
 //var sumBox = $(divObject);

 var sumBox = document.getElementById(divObject);

Obviously this cannot work for both scenarios, without manually commenting and uncommenting the two lines (e.g. on page OpenClinica/SignStudySubject the link “Show events and discrepancy notes” is not functional).

The altered code is first checking to see if divObject is an element or its ID and then is declaring sumBox accordingly.

signstudysubject01
signstudysubject02

### DIFF
--- a/web/src/main/webapp/includes/global_functions_javascript.js
+++ b/web/src/main/webapp/includes/global_functions_javascript.js
@@ -343,7 +343,12 @@ function setNewIconInParentWin(idOfImageElement, imageLocation){
 }
 function showSummaryBox(divObject,parentLinkObj,showText,hideText){
     //var sumBox = $(divObject);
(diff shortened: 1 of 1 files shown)


angood:
Hello,
Thanks for the contribution! We're looking into this and will send you an update soon. We need to create a story for this in Jira and make sure we understand the implementation. We'll get back to you with any questions or feedback. If we're able to accept the pull request we'll let you know what version it will be in. 

Best Regards,
Alicia

PR #195, opened 2015-05-28
Update signStudySubject.jsp

iDrakos:
On pull request #194 I’m proposing an update to the showSummaryBox js function to be able to handle both elements and elemants IDs.

Alternative, if you prefer to keep global_functions_javascript.js intact, you can update the signStudySubject.jsp file and call the function showSummaryBox in the correct way (passing element ID, not the element).

I am also suggesting that the discrepancy notes should be displayed by default, since the information that it contains is vital for many SOP’s. Therefor, link 436 should be changed from:

```
<div id="subjDiscNoteDiv" class="subjDiscNoteDiv" style="display:none">
```

to

```
<div id="subjDiscNoteDiv" class="subjDiscNoteDiv">
```


angood:
Hi @iDrakos  
I agree that it is useful to have this information presented on the signature page as it's quite relevant to the data. We'll look into this and get back to you ASAP. 

Thanks! 
Alicia

PR #222, opened 2015-06-29
OC-5994 Unable to asign user role due to data base column definition for

jkeremian:
user_name has different lengths


kkrumlian:
:+1: Looks good

PR #226, opened 2015-07-01
OC-6496A

jkeremian:
OnSchedule should run with item based rules & rule assignments
Bug fix for multiple trigger rules


kkrumlian:
:+1: Looks good

PR #228, opened 2015-07-06
OC-6528

jkeremian:
Change default time of Run On Schedule to 20:00
build a rest call for the RunTime and updated the rules_template with
comment file to reflect correct time.


core/src/main/java/org/akaza/openclinica/service/JobTriggerService.java
```diff
@@ -131,91 +131,7 @@ public void triggerJob() throws NumberFormatException, ParseException {
 		}
 	}
```
kkrumlian: Hey @jkeremian Did you mean to remove all this code here ?

kkrumlian:
:+1: Looks good

PR #231, opened 2015-07-08
OC-6285: OC 3.4.1 fails to start on fresh install, liquibase exceptio…

sdibona:
…n.  Updated changeset so views will be dropped, then recreated so owner can be updated.


kkrumlian:
:+1: Looks good

PR #245, opened 2015-07-17
OC-828: bugfix (Notes & Discrepancy assigned to me on home page not working)

chaensel:
bugfix for bug OC-828 (https://jira.openclinica.com/browse/OC-828)
If user A (logged in on study-level) assigns a DN to user B (site-level access only), the "Notes & Discrepancy assigned to me:" count is incorrect.


chaensel:
the modification is based on the implementation from ChangeStudyServlet

PR #253, opened 2015-07-24
bugfix for OC-3333 and OC-3106

chaensel:
this fixes the bug OC-3333 and probably the bug OC-3106


chaensel:
Hi Krikor (@kkrumlian)

I barely hoped that this will ever happen but I'm glad that this fix was useful. Is there something I can do to make the pull requests be accepted faster?

Regards,
Christian

PR #258, opened 2015-07-31
Oc 6582a and OC-6583a

jkeremian:
Study designer can configure study and site-specific anonymous URL 
Added session attributed for changes to take into effect when duplicate values are entered.
Also, feedback from demo changes implemented.


kkrumlian:
:+1: Looks good

PR #261, opened 2015-08-04
OC-6624: PersonID subject attribute searchable in Subject Matrix.

toskrip:
This implements the possibility to view/filter/sort PersonID subject attribute in Subject Matrix.


angood:
HI @toskrip , 
How will this present to the user? Will it be another column? If so, will the column always be visible or would it only show on "show more"? 

My feeling is that many of our users do not use Person ID so I wouldn't want it to be shown by default. 

Thanks
Alicia

toskrip:
Yes it is shown as another column in Subject Matrix. One additional column will be visible automatically (without the click on show more). I was thinking of hiding it but I failed to understand how Show More/ Hide logic is implemented in OC (looks like some magic numbers/ hard coded indexes are there). I would be happy if one of your developers can look at it.

PR #264, opened 2015-08-04
OC-6571

jkeremian:
Create OC endpoint to persist follow up 
Created a Rest Call to enter Discrepancy Note for Event fields


kkrumlian:
:+1: Looks good

PR #271, opened 2015-08-10
OC 2893 

rooney:
Attached files are not accessible when the items in the CRF are modified from a different level(study/site) during admin editing


kkrumlian:
:+1: Looks good

PR #273, opened 2015-08-10
OC-6608

sidr23:


web/src/main/resources/org/akaza/openclinica/i18n/words.properties
```diff
@@ -3004,6 +3004,7 @@ false = FALSE
 enabled=Enabled
 disabled=Disabled
 participant_form=Participant Form
+participant_portal= Participant Portal
```
kkrumlian: Can you please check with Alicia and make sure that's what she wants to call it. I think she might want to call it Participate.

PR #275, opened 2015-08-11
OC-6636

jkeremian:
OpenClinica creates API keys and stores them in the DB for new & updated
user accounts


web/src/main/java/org/akaza/openclinica/control/admin/CreateUserAccountServlet.java
```diff
@@ -379,4 +387,20 @@ private void sendNewAccountEmail(UserAccountBean createdUserAccountBean, String
     protected String getAdminServlet() {
         return SecureController.ADMIN_SERVLET_CODE;
     }
+
```
kkrumlian: This code is ok here for now, when we implement UI portion of this story i would move this code to a ApiKeyService

kkrumlian:
:+1: Looks good

PR #294, opened 2015-09-02
OC-6670 Root user cannot log in to study that is made available in Japanese

jkeremian:
Root user cannot log in to study that is made available in Japanese


kkrumlian:
:+1: Looks good

PR #296, opened 2015-09-03
OC-6571 & OC-6468

kkrumlian:
OC-6571: Create OC endpoint to persist follow up
OC-6468: OC Participate - Item Header appears left-aligned with break lines


kkrumlian:
:+1: Looks good

PR #299, opened 2015-09-08
remove null-study check from OpenRosaServices.java

rooney:
revert the NoSuchRequestMethodHandlingException throwing when study is null from OpenRosaServices.getParentStudy()


kkrumlian:
:+1: Looks good

PR #300, opened 2015-09-09
OC-6677  Date of Birth configuration effect on audit log

jkeremian:
Date of Birth configuration effect on audit log


kkrumlian:
:+1: Looks good

PR #303, opened 2015-09-11
OC-6687: Xform CRF upload should not error out because the website attempts to create duplicate OIDs

sdibona:


kkrumlian:
:+1: Looks good

PR #304, opened 2015-09-11
OC-6688: Xform CRF creation should not fail when an item has the same name as the top level tag in the instance

sdibona:


kkrumlian:
:+1: Looks good

kkrumlian:
:+1: Looks good

PR #313, opened 2015-09-23
OC-6727: Participate not loading correct event if there are non-compl…

sdibona:
…eted, non-Participate events scheduled prior to the Participate event


kkrumlian:
:+1: Looks good

PR #323, opened 2015-10-07
OC-6756: Participate Dashboard can load the wrong next event when the…

sdibona:
…re are both manually generated and rule generated events on the same day


kkrumlian:
:+1: Looks good

PR #334, opened 2015-10-19
OC-6776: OpenClinica incorrectly returning 200 HTTP response instead of 201 HTTP response on successful xform submission

sdibona:


kkrumlian:
:+1: Looks good

PR #335, opened 2015-10-20
Remove all artificial file upload size limits by setting fileSizeMax set to Long.MAX_VALUE

mjevans:
This change has been tested against OpneClinica 3.7, and is quite minor and therefore likely to also work with the current head which the original change has been re-based on.

It removes an explicit internal attachment limitation which is inflexible, allowing the tomcat connector properties (maxPostSize and maxSavePostSize), reverse proxy limits, or connection/storage limitations of the involved system(s) to dynamically control the limitation of uploaded file sizes.

As noted in the only changed file, this change is small enough to be classified as a derivative work, which I have performed under the permission of my present employeer during office hours.  As such I have requested and received permission to contribute back to OpenClinica.

* The derivative works are initially licensed by the original modification authors under the same license as the source (in this case the GNU LGPL 2.1) or any greater (future) version there of.
* Upon the first publication under an FSF approved Free Software License of those modifications in the upstream work (EG: OpenClinica) by the owner of that work, said owner has the option to receive full ownership of the modifications that created the derivative work, thus being able to re-license them as they desire for future publications.


mjevans:
a JIRA issue has also been created.

https://jira.openclinica.com/browse/OC-6779

PR #337, opened 2015-10-20
Oc 6585

ivansugi:


core/src/main/java/org/akaza/openclinica/service/rule/expression/ExpressionService.java
```diff
@@ -841,7 +841,8 @@ private String getOidFromExpression(String expression, int patternIndex, int exp
         // int patternIndex = ?;
         if (!match(splitExpression[splitExpression.length - 1 - expressionIndex], pattern[patternIndex])) {
             if (!match(splitExpression[splitExpression.length - 1 - expressionIndex], ruleActionPattern[patternIndex])) {
-                throw new OpenClinicaSystemException("OCRER'R_0019", new String[] { expression });
+                //throw new OpenClinicaSystemException("OCRER'R_0019", new String[] { expression });
+                return expression;
```
kkrumlian: Do you know why we are making this change?

ivansugi:
cause some times when i run a schedule, it gives an exceptions and cause the alicia's error on friday.. but not on all schedules, so when it hits exception, i returns the plain schedule. but i dont know the function of it.

kkrumlian:
:+1: Looks good

PR #338, opened 2015-10-20
OC-3334 MainMenu incorrectly uses full URI

rooney:
MainMenu incorrectly uses full URI for stylesheet may interfere with
some reverse proxy setups.

Instead of using ContextPath better if we use path from sysURL in
datainfo.properties.

We have no idea of how the reverse proxy was set, one thing we know is
that variable sysURL that the sysadmin set to match the reverse proxy
configuration.

Helps with consistency, for example to avoid similar use of urlPrefix
(which value either empty or '../' depending on circumstances) in
navBar.jsp.


kkrumlian:
@rooney Don't think we need this thorough of an implementation so closing for now

PR #339, opened 2015-10-20
OC-6781

jkeremian:
Event Definition API needs to set event ordinals
Updated study_event_definition Ordinals


kkrumlian:
:+1: Looks good

PR #345, opened 2015-10-23
Oc 6780a Missing message in div.alert element when user creates subject and schedules event

jkeremian:


kkrumlian:
Looks good :+1:

PR #347, opened 2015-10-24
Cleanup the gitignore

rooney:


kkrumlian:
@rooney Once OC starts the first time then .config folder is created. So you should change the values in the new .config folder. Let's keep this implementation as is for right now.

PR #351, opened 2015-10-28
OC-6798 API to Create Event won't create repeating group events

jkeremian:
API to Create Event won't create repeating group events
minor change in code to fix this issue


kkrumlian:
:+1: Looks good

PR #357, opened 2015-11-03
OC-6738 Notification action should send email via ocui when Participate is enabled

jkeremian:
Notification action should send email via ocui when Participate is
enabled


kkrumlian:
:+1: Looks good

PR #362, opened 2015-11-04
Oc 6811DEVBUG: Notification Action & Hardcoded Emails

jkeremian:


core/src/main/java/org/akaza/openclinica/domain/rule/action/NotificationActionProcessor.java
```diff
@@ -270,6 +270,10 @@ public void run() {
 
 		} else {
 			pDTO = buildNewPDTO();
+            message = message.replaceAll("\\\\n", "\n");
```
kkrumlian: While you are there can you please remove the empty System.out and format these lines.

kkrumlian:
:+1: Looks good

PR #368, opened 2015-11-09
OC-6817: OpenClinica not throwing correct error when Participate disa…

sdibona:
…bled and an xform is submitted


kkrumlian:
:+1: Looks good

PR #373, opened 2015-11-16
OC-6820 Apply Uniqueness Constraint to study event table

jkeremian:
Apply Uniqueness Constraint to study event table


kkrumlian:
:+1: Looks good

PR #378, opened 2015-11-20
Fix for OC-6061 CRF locked error message

jacobrousseau:
The CRFLocker was not cleared when a session-time out occurs. Tested this fix by changing the session-time in the datainfo.properties to 20 seconds and waiting until the auditLogout method of the OpenClinicaSessionRegistryImpl was triggered. 
* this change was not included in the OpenClinica-ws source because I noticed that the OpenClinicaUsernamePasswordAuthenticationFilter class in the WS-war does not use the CRFLocker.
* the OpenClinicaSessionRegistryImpl in the WS does not retrieve the user information from LDAP, the Web however it does use LDAP.


kkrumlian:
Thank you @jacobrousseau for the contribution will push it in 3.9.

kkrumlian:
:+1: Looks good

PR #379, opened 2015-11-20
Bugfix, first-time run (after installation) can't create openclinica.config

rooney:
Installing OpenClinica to tomcat should've created openclinica.config
directory in $CATALINA_HOME if there isn't one already.


kkrumlian:
@rooney So if i understand this correctly - 
* If URL, Hibernate Dialect, Driver don't have values then openclinica.config is not being create?

Krikor

rooney:
@kkrumlian Yes. 
Or, at least that should be documented that those configs must have values

kkrumlian:
No need to push this

PR #380, opened 2015-11-20
OC 6720 null-dates fix

rooney:


core/src/main/resources/properties/study_event_dao.xml
```diff
@@ -193,9 +193,10 @@
         <name>findAllBySubjectIdOrdered</name>
         <sql>
 			SELECT se.*
-            FROM study_event se, study_event_definition sed
+            FROM study_event se, study_event_definition sed, study_subject ss
             WHERE se.study_event_definition_id = sed.study_event_definition_id
-            	AND se.study_subject_id=?
```
kkrumlian: I don't think we need a fix here. Steve made a fix in the code.

kkrumlian:
No action required

PR #383, opened 2015-11-24
OC-6777: Participate/Enketo *edit* form submissions do not handle repeating group row deletions correctly 

sdibona:


kkrumlian:
Looks good. I will open up a new story to address show/hide action behavior

kkrumlian:
:+1: Looks good

PR #385, opened 2015-11-30
OC-6720 not-displaying-a-date bugfix

rooney:
Krikor reported a bug via Hipchat that dates not displayed.
This is the bugfix for that.


kkrumlian:
This doesn't make sense to me. The query that this is replacing returns the same result. @rooney Let's chat tomorrow

kkrumlian:
Ok i'm going to close this - The implementation that we put in place is here: https://github.com/OpenClinica/OpenClinica/pull/388

PR #398, opened 2015-12-07
OC-6870: Submitting Participate data for an Excel-based CRF does not …

sdibona:
…set a Schedule Study Event to Data Entry Started status


kkrumlian:
:+1: Looks good

PR #400, opened 2015-12-07
OC-6812

jkeremian:
Creating a site or an event generates an Oops Error when a study is
enabled for Participate but OCUI is down


kkrumlian:
@jkeremian I like the refactoring. Moving all the code into SecureController allows us not to duplicate that code everywhere.

kkrumlian:
:+1: Looks good

PR #401, opened 2015-12-07
OC-6536 OpenClinica allows you to upload a rule when all the run on parameters are set to 'FALSE'

jkeremian:
OpenClinica allows you to upload a rule when all the run on parameters
are set to 'FALSE'


kkrumlian:
:+1: Looks good

PR #402, opened 2015-12-07
OC-6483 Notification Rule with more than 255 characters in <Message> and <Subject> elements generate Oops Error

jkeremian:
Notification Rule with more than 255 characters in <Message> and
<Subject> elements generate Oops Error


kkrumlian:
:+1: Looks good

PR #405, opened 2015-12-10
OC-2183 Bugfix, filters at Administer Subjects

rooney:
The subject table in Tasks->Administer->Subjects has its filter stopped working.


kkrumlian:
👍 Looks good

PR #406, opened 2015-12-10
OC-6835: Xform repeating group edit additions must work when jr:templ…

sdibona:
…ate not defined in instance for that repeating group.


kkrumlian:
:+1: Looks good

PR #409, opened 2015-12-10
Oc 2047 Rules not working when absolute paths (versus relative paths) are used in the expression or action

jkeremian:
Rules not working when absolute paths (versus relative paths) are used in the expression or action


kkrumlian:
:+1: Looks good

PR #410, opened 2015-12-11
OC-6877 DEV BUG: Valid Event Action rule is getting a NullPointer Exception when attempting to validate or save in Rule Designer

jkeremian:
DEV BUG: Valid Event Action rule is getting a NullPointer Exception when
attempting to validate or save in Rule Designer


kkrumlian:
:+1: Looks good

PR #411, opened 2015-12-11
OC-6649 Participate Study Configuration Should not Appear on the View Study Page if Participant Portal is not Enabled 

jkeremian:
Participate Study Configuration Should not Appear on the View Study Page
if Participant Portal is not Enabled


kkrumlian:
:+1: Looks good

PR #413, opened 2015-12-15
Oc 6787

ivansugi:
is this good?


web/target/OpenClinica-web-MAINLINE-SNAPSHOT/WEB-INF/jsp/submit/import.jsp
```diff
@@ -0,0 +1,105 @@
+<%@ page contentType="text/html; charset=UTF-8" %>
```
kkrumlian: @ivansugi Please remove this file. This is in target folder and it shouldn't be comitted

PR #414, opened 2015-12-15
OC-6867: Quick Editing hangs

sdibona:


jbonica:
Looks good -- per Krikor.

PR #415, opened 2015-12-15
OC-6483 Added validation to avoid oops page on upload for characters/ bytes over specified #

jkeremian:
Notification Rule with more than 255 characters in <Message> and
<Subject> elements generate Oops Error


kkrumlian:
:+1: Looks good

PR #417, opened 2015-12-17
OC-6883 Running InsertAction rules in batch on a CRF that has been marked complete

jkeremian:
Running InsertAction rules in batch on a CRF that has been marked
complete


kkrumlian:
:+1: Looks good

PR #419, opened 2015-12-17
OC-6836: Repeating group rows deleted from xforms in Participate will…

sdibona:
… be editable in OpenClinica


kkrumlian:
:+1: Looks good

PR #442, opened 2015-12-29
OC-6900 Bugfix, disable Participate in Build Study

rooney:
Issue description:

> Disabling Participate under Build Study page by clicking [X] does change the
> status to Disabled while it does not change the actual status in OCUI, and
> the status is back into Enabled when user log back in.


kkrumlian:
:+1: Looks good

PR #451, opened 2016-01-05
OC-2047 Absolute vs Relative Rule Expression

jkeremian:
Rules not working when absolute paths (versus relative paths) are used
in the expression or action


kkrumlian:
:+1: Looks good

PR #453, opened 2016-01-06
OC-6913 DEV BUG: Remove exception from register randomize module

jkeremian:
DEV BUG: Remove exception from register randomize module


kkrumlian:
:+1: Looks good

PR #454, opened 2016-01-07
OC-1176 Validating rule_expression_value length at rule upload to OC

jkeremian:
Extend rule_expression.value data type beyond 1025 characters


kkrumlian:
:+1: Looks good

PR #459, opened 2016-01-13
OC-6847 Randomize Module

jkeremian:
At Randomize Access Request where OCUI is not running , the user gets
email notification for the request.


kkrumlian:
:+1: Looks good

PR #462, opened 2016-01-14
OC-6925 Validate item belongs to item group in rule experssions

jkeremian:
Validate item belongs to item group in rule experssions


kkrumlian:
:+1: Looks good

PR #472, opened 2016-02-13
OC-6990 Show offline option(s) in all representations of study metadata

jkeremian:
Show offline option(s) in all representations of study metadata


kkrumlian:
:+1: Looks good

PR #473, opened 2016-02-16
OC-6990 Show offline option(s) in all representations of study metadata

jkeremian:
Show offline option(s) in all representations of study metadata
Fixed a minor bug where the study_event is non repeating


kkrumlian:
:+1: Looks good

PR #477, opened 2016-02-17
OC-7012 fix for event definition page table borders

jkeremian:
fix for event definition page table borders


kkrumlian:
:+1: Looks good

PR #482, opened 2016-02-22
Make OCUI page offline

rooney:
OC-7010


rooney:
Not needed because Steve will update OC's API

PR #486, opened 2016-02-23
Oc 7013 Subject can capture an image in the questionnaire 

jkeremian:
Subject can capture an image in the questionnaire


kkrumlian:
:+1: Looks good

PR #488, opened 2016-02-24
OC-7013 Subject can capture an image in the questionnaire 

jkeremian:
Subject can capture an image in the questionnaire 
-Fix for uploading any image file supported by enketto
-added validation at upload time for non image Upload mediatypes
-also fix for view_event_definition page table mis-aligned


kkrumlian:
:+1: Looks good

PR #496, opened 2016-03-01
OC-3333 The rule does not fired with <Run InitialDataEntry="false" DoubleDataEntry="true" > setting 

jkeremian:
The rule does not fired with <Run InitialDataEntry="false"
DoubleDataEntry="true" > setting


kkrumlian:
:+1: Looks good

PR #498, opened 2016-03-02
OC-6932 Blocking Monitoring related Rest Calls for non authenticated users

jkeremian:
Blocking Rest Calls for non authenticated users


kkrumlian:
:+1: Looks good

PR #499, opened 2016-03-02
OC-7034: For new xform submission code, if specified Study Subject ID does not exist, create it rather than doing an anonymous submission

sdibona:


kkrumlian:
:+1: Looks good

PR #501, opened 2016-03-03
OC-7060 Define reserved term for confirming study subject ID

jkeremian:
Define reserved term for confirming study subject ID


kkrumlian:
:+1: Looks good

PR #503, opened 2016-03-04
OC-6587 Item validation for DATE/INT treats whitespace as empty/null but saves the value anyway

jkeremian:
Item validation for DATE/INT treats whitespace as empty/null but saves
the value anyway


kkrumlian:
:+1: Looks good

PR #507, opened 2016-03-07
OC-7074: Xform Submission Service should skip items it can't identify rather than aborting the data import

sdibona:


kkrumlian:
:+1: Looks good

PR #510, opened 2016-03-08
OC-7066 For an event to have an anonymous form it must be repeating

jkeremian:
For an event to have an anonymous form it must be repeating


kkrumlian:
:+1: Looks good

PR #513, opened 2016-03-09
OC-6711 OC does not record correct user for CRF Version Migration in Audit Logs

jkeremian:
OC does not record correct user for CRF Version Migration in Audit Logs


kkrumlian:
:+1: Looks good

PR #514, opened 2016-03-09
OC-5918 Rule ShowAction/HideAction not behaving properly after CRF migration

jkeremian:
Rule ShowAction/HideAction not behaving properly after CRF migration


kkrumlian:
:+1: Looks good

PR #515, opened 2016-03-10
OC-7094: Pform submissions that fail with non-validation errors (unhandled exceptions) are sending a success response to Enketo

sdibona:


kkrumlian:
:+1: Looks good

PR #517, opened 2016-03-11
OC-7100 Remove sort link in the manage event definition page

jkeremian:
Remove sort link in the manage event definition page


kkrumlian:
:+1: Looks good

PR #518, opened 2016-03-14
OC-7095 API must not include secondaryProtocolID

jkeremian:
API must not include secondaryProtocolID


kkrumlian:
:+1: Looks good

PR #523, opened 2016-03-16
OC-7031: Mandatory attribute in study metadata not recorded properly …

sdibona:
…for required items in XForms


kkrumlian:
:+1: Looks good

PR #524, opened 2016-03-16
OC-7106 OC3.10 Dev Bug: Event Definition page has extra empty <td> tag that stops UPOQ to work

jkeremian:
OC3.10 Dev Bug: Event Definition page has extra empty <td> tag that
stops UPOQ to work


kkrumlian:
:+1: Looks good

PR #525, opened 2016-03-16
OC-6879 Removing (Pausing) an Export Job Results in the Job Still Running at the Scheduled Time

jkeremian:
Removing (Pausing) an Export Job Results in the Job Still Running at the
Scheduled Time


kkrumlian:
:+1: Looks good

PR #526, opened 2016-03-16
OC-6802 Update documentation on text and text area character limit

jkeremian:
Update documentation on text and text area character limit


kkrumlian:
:+1: Looks good

PR #528, opened 2016-03-17
for OC-7086 Load offline dashboard

rooney:
return name and description of crf along with enketo url


kkrumlian:
:+1: Looks good

PR #529, opened 2016-03-17
OC-7083 CRF can be marked complete without all required fields being completed

jkeremian:
CRF can be marked complete without all required fields being completed


kkrumlian:
:+1: Looks good

PR #531, opened 2016-03-21
OC-7006 clinical research coordinator can view offline form

rooney:
Category: feature request

As a clinical research coordinator that is preparing a tablet to be used in an offline setting
I want to be able to access the necessary forms that will be used offline

AC1: The URLs must be accessible to the CRC

AC2: The URLs must be hyperlinked so that they can easily launch the form in their browser

AC3: Each URL must open in a new window so that it is easy to open more than one form for offline use


kkrumlian:
This should be closed

PR #536, opened 2016-03-23
OC-7051 end point for StudyUserRoles

rooney:


kkrumlian:
Not sure this is implemented the right way. It needs to be a new REST endpoint. Closing for now

PR #544, opened 2016-03-29
OC-7015 Add margins to printable CRFs

rooney:
The look and readability of printable CRFs would be improved with some
whitespace in the margins.


kkrumlian:
:+1: Looks good

PR #548, opened 2016-03-31
OC-7118 CRC can see access code after connecting a participant

jkeremian:
CRC can see access code after connecting a participant


kkrumlian:
:+1: Looks good

PR #551, opened 2016-03-31
OC-2049 Hidden items in the repeating group causes undesirable forced reason for change during administrative editing (no data update)

jkeremian:
Hidden items in the repeating group causes undesirable forced reason for
change during administrative editing (no data update)


kkrumlian:
:+1: Looks good

PR #554, opened 2016-04-05
OC-7115 OpenClinica adding incorrect audit entries for repeating groups submitted through a Participant Form

rooney:


kkrumlian:
Implemented different way

PR #555, opened 2016-04-05
Update audit user activity to log when access code viewed by crc user (OC-7118)

rooney:


core/src/main/resources/migration/3.10/2016-04-05-OC-7118.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
kkrumlian: @jkeremian Is this something that is needed ?

kkrumlian:
👍 Closing

PR #559, opened 2016-04-08
OC-7137 Open Participant form in OpenClinica but not saving makes form invalid for participant

jkeremian:
Open Participant form in OpenClinica but not saving makes form invalid
for participant


kkrumlian:
👍 Looks good

PR #562, opened 2016-04-12
OC-7119 No properties for new items after CRF version migration

jkeremian:
No properties for new items after CRF version migration


kkrumlian:
👍 Looks good

PR #564, opened 2016-04-13
OC-7150 OpenClinica username length requirement (min. 5 chars) might conflict with LDAP

jkeremian:
OpenClinica username length requirement (min. 5 chars) might conflict
with LDAP


kkrumlian:
👍 Looks good

PR #568, opened 2016-04-26
OC-7160 REST Endpoint that produces OC specific configuration information

jkeremian:
REST Endpoint that produces OC specific configuration information


kkrumlian:
👍 Looks good

PR #579, opened 2016-05-16
Fix for rules not executing properly (When participate submits & when…

kkrumlian:
… rules run on schedule)


kkrumlian:
👍 Looks good

PR #583, opened 2016-05-31
3.6

jacobrousseau:
Hi Krikor,

after a couple of weeks I've found some time to finish work on a proposition we had a discussion about on the OpenClinica forum. It was about adding extra information on the events to enable Webservice clients to determine if events exist and what the CRF / CRF-version state was. I followed Tomas' suggestion to do it on the listAllByStudy operation because it allready contained a lot of the information. The changes were done on the 3.6 branch because that's the version we are working with, I trust that merging them into the main-branch will be relatively easy. Also I've made all the fields optional to avoid other clients being forced to change their code. As you remarked a this would not require a new XSD-version. Could you let us know if you consider this modification is OK and if it will be added to the main-branch?
regards, Jacob

### DIFF
--- a/ws/src/main/java/org/akaza/openclinica/ws/StudySubjectEndpoint.java
+++ b/ws/src/main/java/org/akaza/openclinica/ws/StudySubjectEndpoint.java
@@ -237,7 +237,10 @@ public class StudySubjectEndpoint {
             if (studyRefOutput == null) {
                 StudyBean siteStudy = (StudyBean) getStudyDao().findByPK(studyIDSubject);
                 studyRefOutput = new StudyRefType();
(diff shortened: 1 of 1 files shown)


toskrip:
Hey Jacob, thanks a lot for this. Next week I will put it to my testing environment just to verify whether it has some side effects for some of my existing clients. This would indeed simplify my clients codebase. I will let you know if I find some issue or possible enhancement. best, Tomas

jacobrousseau:
Hi Tomas,


no problem. Please let us know if you have any problems with your clients with namespaces etc. I'll keep an eye on the OpenClinica-forum.


regards, Jacob

________________________________

PR #594, opened 2016-07-14
Oc 7292 New Bug in 3.11 got introduced:Hidden required fields prevent CRF from being marked compete

jkeremian:


kkrumlian:
Merging this as part of a separate pull request

PR #604, opened 2016-07-20
OC-7282: Add support for user lists for Enketo Discrepancy Notes

jkeremian:


sdibona:
AC changed after code checked in.  Code does not fulfill all AC.

PR #606, opened 2016-07-23
OC-7343 Run batch CRF version migration as an api

jkeremian:
Run batch CRF version migration as an api


kkrumlian:
👍 Looks good

PR #607, opened 2016-07-25
OC-7292

kkrumlian:
New Bug in 3.11 got introduced:Hidden required fields prevent CRF from
being marked compete

In the code , I have include Group check (Show_group) in addition to
Item check for (show_item)


kkrumlian:
👍 Looks good

PR #624, opened 2016-08-12
Oc 5891 OpenClinica silently fails to create a DN when the rule message is greater than 255 characters long

jkeremian:
OpenClinica silently fails to create a DN when the rule message is greater than 255 characters long


yogi-oc:
Looks good.

PR #625, opened 2016-08-12
Oc 7200 Capitalize the word "data" and "syntax" in "SPSS data and syntax" so it reads "SPSS Data and Syntax"?

jkeremian:
Capitalize the word "data" and "syntax" in "SPSS data and syntax" so it reads "SPSS Data and Syntax"?


kkrumlian:
👍 Looks good

Issue #527, opened 2016-03-17
Maven can't find spring-security-oauth 1.0.0.M2


laowood:
The Central Repository can't find this dependency
`
<dependency>
	<groupId>org.springframework.security.oauth</groupId>
	<artifactId>spring-security-oauth</artifactId>
	<version>1.0.0.M2</version>
</dependency>
`

I Changed to 

`
<dependency>
	<groupId>org.codehaus.spring-security-oauth</groupId>
	<artifactId>spring-security-oauth</artifactId>
	<version>3.19.SS3</version>
</dependency>
`

![qq20160317-0 2x](https://cloud.githubusercontent.com/assets/4517783/13837171/730d135a-ec43-11e5-92d9-3ff79d119821.png)
![qq20160317-1 2x](https://cloud.githubusercontent.com/assets/4517783/13837172/73105c4a-ec43-11e5-8f49-7dfe8cf3d983.png)

changjiasheng:
you can chang the pom.xml like this 
` <dependency> <groupId>org.springframework.security.oauth2</groupId> <artifactId>spring-security-oauth</artifactId> <version>1.0.0.M3</version> </dependency> `

after this ,you will be fine

liyanqing1992:
but there is still some problem left after using the way metioned by @changjiasheng  
![image](https://cloud.githubusercontent.com/assets/9292933/15567842/235efac2-235b-11e6-9d71-7ece77f44f01.png)

PR #634, opened 2016-08-18
Changed version of spring-security-oauth to 1.0.0.M3 

sjuan76:
As stated by https://github.com/OpenClinica/OpenClinica/issues/527


yogi-oc:
We cannot change the spring-security-oauth to 1.0.0.M3 because it breaks the Rule Designer. It will be changed in OC 4.0 release

PR #637, opened 2016-08-22
OC-7271 Fix issue with AddNewSubject page

jkeremian:
Fix issue with AddNewSubject page
Added validation at subject create and update time not to allow saving
html less than or greater than elements


kkrumlian:
👍 Looks good

PR #639, opened 2016-08-23
Oc 7396 Secured API allowing access with an empty String API key when db has Anonymous Participants

jkeremian:


kkrumlian:
👍 Looks good

PR #640, opened 2016-08-24
Some .gitignore improvements

aistomin:
I just noticed that Idea generated `stormpath/Stormpath.iml` file which was not excluded from version control. Then I started fixing that and made some cleaning in `.gitignore`. Maybe those changes will be useful for you as well. 
I've done the following changes:
- instead of mentioning file by file I've just excluded all the *.iml files from version control;
- moved all the IntelliJ related lines to special block(similar to Git and NetBeans blocks).
Please, accept this pull request if you think that it makes sense.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ release.properties
 .classpath
 .project
 .settings/
-.idea/
 web/src/main/resources/datainfo.properties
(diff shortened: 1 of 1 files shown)


yogi-oc:
Thank you for your contribution.

PR #641, opened 2016-08-24
OC-6997: i18n: Rule against Date won't triger if Date Format is European

azfalot:
Use locale in date conversion when running rules in CRF from and rule test page.


yogi-oc:
The date format has to be specified in the properties file.

gabrielgt:
The date format is already correctly specified in both files, format.properties and format_es.properties:

"date_format_string = dd-MMM-yyyy"

But the problem are those months with different abreviature in English and Spanish, for example:

January (Jan) - Enero (Ene): it does not work
February (Feb) - Febrero (Feb): it works

To convert "Ene" in 01 is necesary to use the locale and the format.properties file does not have it.

Have you read OC-6997 issue in Jira? 
If so, please, Could you explain to us how should we solve the problem?

Thanks for your support!

yogi-oc:
I've incorporated these changes in the 3.12 branch.
Thanks for your contribution._

PR #642, opened 2016-08-24
OC-3.12 Added missing import to studySubjectDAO file

jkeremian:
Added missing import to studySubjectDAO file which was causing Jenkin fail


kkrumlian:
👍 Looks good

PR #649, opened 2016-08-30
OC-7406 Devbug: Batch CRF v. Migration API with Removed/Archived CRFs

jkeremian:
Devbug: Batch CRF v. Migration API with Removed/Archived CRFs
added formatting


yogi-oc:
I understand the formatting doesn't look right but it looks OK in the committed file, so not sure what the problem is. We'll deal with this later.

PR #660, opened 2016-09-09
Master jira oc 7427

sjuan76:
In order comparation of the SHA-1 of the password case insensitive, the incoming SHA-1 is passed to lowercase.

Of course, password check through the web remains case sensitive (in case you were wondering) :-D


sjuan76:
It seems that the pull request added more files than the one I changed, closing it until I get to find how to send only the relevant commit.

PR #662, opened 2016-09-14
Oc 7417 New OpenClinica Extensions for Import ODM 

jkeremian:


core/src/main/java/org/akaza/openclinica/dao/hibernate/EventDefinitionCrfDao.java
```diff
@@ -1,46 +1,58 @@
...
+    Class<EventDefinitionCrf> domainClass() {
+        // TODO Auto-generated method stub
+        return EventDefinitionCrf.class;
+    }
+
+    @SuppressWarnings("unchecked")
+    public List<EventDefinitionCrf> findByStudyEventDefinitionId(int studyEventDefinitionId) {
+        String query = "from " + getDomainClassName()
+                + " event_definition_crf where event_definition_crf.studyEventDefinition.studyEventDefinitionId = :studyeventdefinitionid";
+        org.hibernate.Query q = getCurrentSession().createQuery(query);
```
yogi-oc: why do we need to spell out the class path here? "org.hibernate.Query"
Can you make it to read just Query?

PR #701, opened 2016-11-23
OC-7512: Set up Walkme self-hosting in OCBridge

kkrumlian:


kkrumlian:
👍 Looks good after code review - will update plugin version in a subsequent commit

PR #713, opened 2016-12-16
OC-7408

jkeremian:
Support field submission for non-repeating/repeating groups with new group data model including saving queries.


kkrumlian:
👍 Looks good

PR #716, opened 2016-12-28
OC-7664

GerbenRienk:
Pull request for OC-7664: javascript doesn't work when used in DDE.
This is because attribute  escapeXml="false" is not in the tag th, where it does exists in initialDataEntry


GerbenRienk:
Please re-open

PR #722, opened 2017-01-12
Devbug: Upload XLSform, Form manager sending actual form name instead…

kkrumlian:
Devbug: Upload XLSform, Form manager sending actual form name instead of form


kkrumlian:
👍

PR #724, opened 2017-01-13
Devbug: Few jersey based REST endpoints not wired correctly

kkrumlian:


kkrumlian:
Reverting few changes. Jersey endpoints being used also from Spring controllers hence creating weird wiring issues and strategies.

PR #734, opened 2017-01-23
Oc 7764 Deleting an Event_crf throws an Oops error page at confirmation page 

jkeremian:


web/src/main/java/org/akaza/openclinica/controller/openrosa/OpenRosaSubmissionController.java
```diff
@@ -251,7 +251,9 @@
 
         int count = 0;
         for (EventCrf evCrf : eventCrfs) {
-            if (evCrf.getStatusId() == org.akaza.openclinica.domain.Status.UNAVAILABLE.getCode()) {
+            if (evCrf.getStatusId() == org.akaza.openclinica.domain.Status.UNAVAILABLE.getCode()
```
yogi-oc: Would be useful to import this org.akaza.openclinica.domain.Status statically so we don't use long names

jkeremian: There are 2 different imports for Status class in OC.
 org.akaza.openclinica.bean.core.Status;
 org.akaza.openclinica.domain.Status
The first one has already been imported and in use in another method and I need to use
2nd option with hibernate.

yogi-oc: Would it be possible to do something like this:
org.akaza.openclinica.domain.Status as HibernateStatus

jkeremian: Yogi,
As a result of my investigation , there is no import aliasing mechanism in Java. You cannot import two classes with the same name and use both of them unqualified.
Import one class and use the fully qualified name for the other one within the code. You can do aliasing with Groovy. thanks

PR #736, opened 2017-02-01
OC-7773: DEVBUG: Intermittently Subject Matrix Connect Participant ic…

sdibona:
…on makes stuck at OCUI search window.


jkeremian:
Looks good to me

PR #739, opened 2017-02-02
3.13 oc 7568

hieudd:
Fix i18n for @sdibona comments

### DIFF
--- a/core/src/main/resources/properties/useraccount_dao.xml
+++ b/core/src/main/resources/properties/useraccount_dao.xml
@@ -204,6 +204,34 @@
 			ORDER BY ua.date_created ASC
 		</sql>
 	</query>
+	<query>
+		<name>findAllAssignedUsersByStudy</name>
+		<sql>
+			SELECT
+			ua.user_name
+			, ua.first_name
+			, ua.last_name
+			, sur.role_name
+			, sur.study_id
+			, sur.status_id
+			, sur.date_updated
+			, sur.update_id
+			, s.name
+			, s.parent_study_id
+			, ua.user_id
+			FROM
+			user_account ua
+			, study_user_role sur
+			, study s
+			WHERE ua.user_name=sur.user_name
+			AND (sur.study_id=s.study_id)
+			AND ua.status_id=1
+			AND sur.status_id!=0
+			AND (s.study_id=? OR s.parent_study_id=?)
+			AND  ua.user_name NOT LIKE '%.SS_%'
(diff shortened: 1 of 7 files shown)


sdibona:
Hello,

Thank you for re-creating this pull request against the 3.13 release.  I am reviewing the code changes and will provide an update tomorrow with any suggestions or comments.

Thank you,
-Steve DiBona

sdibona:
Code review in progress.  Will wrap it up tomorrow.

As noted in the other pull requests, I wanted to let you know that due to a tight release deadline there is not enough time to adequately test this change (an automated test needs to be created) before we need to be ready for release. So we will, unfortunately, not be able to include it as part of 3.13. However we are very much planning on including it in an upcoming release. I will provide further updates when more information becomes available.

Please let me know if you have any questions.

Thank you,
Steve DiBona

hieudd:
Thank Steve for your response.

core/src/main/resources/properties/useraccount_dao.xml
```diff
@@ -204,6 +204,34 @@
...
+			, s.parent_study_id
+			, ua.user_id
+			FROM
+			user_account ua
+			, study_user_role sur
+			, study s
+			WHERE ua.user_name=sur.user_name
+			AND (sur.study_id=s.study_id)
+			AND ua.status_id=1
+			AND sur.status_id!=0
```
sdibona: We only want roles in status of 1 (available), 5 (removed), 7 (auto-removed) so this should filter on those roles:

AND sur.status_id IN (1,5,7)

sdibona:
Also this change re-enables an old servlet, RestoreStudyUserRoleServlet.java which is fine.  However this servlet sends out 3 emails every time a role is restored, which is not something we want to do.  So please remove the sendEmail(...) method, but move the portion building the body of the email to the parent method (processRequest()) and use that for the addPageMessage(...) call at line 88.

String restoreMessage =
            u.getFirstName() + " " + u.getLastName() + "(" + resword.getString("username") + ": " + u.getName() + ") "
                + respage.getString("has_been_restored_to_the_study_site") + " " + study.getName() + respage.getString("with_the_role") + " "
                + sub.getRoleName() + ". " + respage.getString("the_user_is_now_be_able_to_access_study");
addPageMessage(restoreMessage);

sdibona:
Merging into OpenClinica:3.14-pullReq739 branch.  I will implement the suggested changes and open a new pull request into 3.14 so that the changes can be merged into the next release today.

PR #742, opened 2017-02-06
OC-7809 Form layouts

jkeremian:
This full request bunch of stories associated with crf version changes using form layout.
It also includes odm metadata and clinical data changes.


kkrumlian:
👍 Looks good

PR #743, opened 2017-02-07
OC-7794 As a protocol manager I want to get the protocol's metadata

jkeremian:
Fix a dev bug issue with Offline Configuration Parameter in Protocol
Metadata


yogi-oc:
looks good.

PR #758, opened 2017-02-15
OC-7762: Wrong question numbering in Randomization API

sdibona:


sdibona:
Krikor reviewed this change.  Looks good.

PR #759, opened 2017-02-16
Fix for OC-7400: "E-mail address entry does not remove spaces". Also …

jacobrousseau:
Hi Steve and or Krikor,
here is an additional fix we have introduced in our fork. When using only the year-of-birth as the date-of-birth of a subject, then the year is not exported to SPSS. I've fixed the 3 XSLT involved.

The orignal PR message was:
here is a PR for OC7400. We hope it will be included in the master branch.

regards, Jacob


jacobrousseau:
Hi Krikor, here is a PR for 2 problems we've recently found in OpenClinica. See the commit comments for the OC JIRA-ID's. Hopefully you'll merge this PR after a review. 
have a nice day, 
regards, Jacob

jacobrousseau:
Hi @kkrumlian ,
included in this PR is an additional fix we have introduced in our fork. When using only the year-of-birth as the date-of-birth of a subject, then the year is not exported to SPSS. I've fixed the 3 XSLT involved.

regards Jacob

PR #765, opened 2017-03-01
OC-7880 Can't open form if item group is not defined for a read-only element 

jkeremian:


core/src/main/java/org/akaza/openclinica/service/crfdata/XformMetaDataService.java
```diff
@@ -530,7 +526,8 @@ public XformContainer parseInstance(String xform, Errors errors, Html html, Stri
             List<XformItem> xformItems = new ArrayList<>();
             int itemOrderInForm = 0;
             for (Bind bd : html.getHead().getModel().getBind()) {
-                if (!bd.getNodeSet().endsWith("/meta/instanceID") && !bodyGroupPaths.contains(bd.getNodeSet())) {
+                if (!bd.getNodeSet().endsWith("/meta/instanceID") && !bodyGroupPaths.contains(bd.getNodeSet()) && (bd.getReadOnly() == null
```
yogi-oc: Can you propose how to simply this condition? It is hard to understand

PR #766, opened 2017-03-01
Oc 7872  OpenClinica:FormLayoutOID attribute is missing from FormData node in Clinical Data XML

jkeremian:


core/src/main/java/org/akaza/openclinica/service/crfdata/XformMetaDataService.java
```diff
@@ -530,7 +526,8 @@ public XformContainer parseInstance(String xform, Errors errors, Html html, Stri
             List<XformItem> xformItems = new ArrayList<>();
             int itemOrderInForm = 0;
             for (Bind bd : html.getHead().getModel().getBind()) {
-                if (!bd.getNodeSet().endsWith("/meta/instanceID") && !bodyGroupPaths.contains(bd.getNodeSet())) {
+                if (!bd.getNodeSet().endsWith("/meta/instanceID") && !bodyGroupPaths.contains(bd.getNodeSet()) && (bd.getReadOnly() == null
```
yogi-oc: Is it possible to rearrange this so it is easier to understand?

web/src/main/java/org/akaza/openclinica/web/pform/OpenRosaServices.java
```diff
@@ -326,8 +327,10 @@ public XFormList getForm(@Context HttpServletRequest request, @Context HttpServl
                 xform = queryForm.decorate(xformParserHelper);
             }
 
-            form.setHash(DigestUtils.md5Hex(xform));
-            // form.setHash(DigestUtils.md5Hex(String.valueOf(cal.getTimeInMillis())));
+            // form.setHash(DigestUtils.md5Hex(xform));
```
yogi-oc: Please remove it if not needed

jkeremian: I will remove it the commented line.

core/src/main/java/org/akaza/openclinica/service/crfdata/XformMetaDataService.java
jkeremian: I will try to see if possible.   In https://jira.openclinica.com/browse/OC-7690
There's an AC1: 'CRF elements can be defined as read-only = yes and calculation attribute = non-null to make them read-only calculated elements'. 
It's a complicated logic. I have to consider also negative logic too. In case where the readonly is not specified , or if it is specified then make sure it is not set to true.

PR #770, opened 2017-03-08
OC-7913 Query Widget losing Type column and Time when re-opening form

jkeremian:
Query Widget losing Type column and Time when re-opening form

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
+++ b/core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
@@ -64,6 +64,7 @@ import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.dbcp.BasicDataSource;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.exception.ExceptionUtils;
+import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
```diff
@@ -248,9 +250,16 @@ public QueriesBean buildQueryElement(ItemData itemdata) {
             }
             query.setComment(dn.getDetailedNotes());
             query.setStatus(dn.getResolutionStatus().getName().toLowerCase());
-            query.setDate_time(dn.getDateCreated().toString());
+
```
yogi-oc: joda.time is in maintenance mode. Please use java.time
For more info look at this link:
http://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html

jkeremian: Enketo expects the following date/timestamp format to work:2017-03-07 17:58:43 -05:00 , otherwise , it will not work.
The date/time format coming from Postgres db is as such:     2017-03-07 17:58:15.068-05
I can't find in java.time any function that will do the trick to convert to the desired format.
 The only one I found was the joda.time (DateTime) function that worked with me. 
I'm running out of options.Thanks

yogi-oc: Use this:
		LocalDateTime.fromDateFields(dn.getDateCreated())

yogi-oc:
Good job.

PR #776, opened 2017-03-17
OC-7939 Completed forms remain completed when data is modified

jkeremian:
Completed forms remain completed when data is modified


yogi-oc:
thanks

PR #780, opened 2017-03-21
Oc 7941 CRF Version Metadata displaying duplicate items when two Layout Versions exist for CRF

jkeremian:


core/src/main/resources/properties/item_dao.xml
```diff
@@ -112,7 +112,7 @@
 	</query>
 	<query>
 		<name>findAllItemsByVersionId</name>
-		<sql>select i.* from versioning_map v, item i 
+		<sql>select distinct i.* from versioning_map v, item i
```
yogi-oc: Do we need to add another table in the query? I am trying to understand why "distinct" is used.

web/src/main/java/org/akaza/openclinica/control/admin/CreateXformCRFVersionServlet.java
```diff
@@ -41,6 +42,7 @@
     Locale locale;
     FileUploadHelper uploadHelper = new FileUploadHelper();
     public final String FM_BASEURL = "http://fm.openclinica.info:8080/api/protocol/";
+    // public final String FM_BASEURL = "http://oc.local:8090/api/protocol/";
```
yogi-oc: you can remove this if not needed.

jkeremian: I need that url for  debuging my local Form Manager application.

core/src/main/resources/properties/item_dao.xml
jkeremian: Due to recent changes to crf_version and  form_layout , where we use the same crf_version against all the from_layouts within CRF, caused the versioning_map table to have multiple [item and crf_version combination] records for each form_layout within CRF.  For this reason, the page was loading duplicate records. By adding distinct to the sql, I'm able to get unique records.

yogi-oc:
great responses to my comments.

PR #781, opened 2017-03-21
Oc 7921 History of Data Change must be displayed in Query Widget after user closes and reopens form

jkeremian:


core/src/main/resources/properties/item_dao.xml
```diff
@@ -112,7 +112,7 @@
 	</query>
 	<query>
 		<name>findAllItemsByVersionId</name>
-		<sql>select i.* from versioning_map v, item i 
+		<sql>select distinct i.* from versioning_map v, item i
```
yogi-oc: same question as the previous pull request.

PR #782, opened 2017-03-21
OC-7943 OCForm is not rendering saved selections in a checkbox when form is closed and re-opened

jkeremian:


kkrumlian:
👍 Looks good

PR #790, opened 2017-04-04
OC-7968 OCForm not assigning Auto-generated Query to data entry user 

jkeremian:
OCForm not assigning Auto-generated Query to data entry user


yogi-oc:
Looks good.

PR #808, opened 2017-05-08
OC-8055 Oops Error occurs when attempting to view metadata of a second CRF AFTER uploading the first CRF

jkeremian:
Oops Error occurs when attempting to view metadata of a second CRF AFTER
uploading the first CRF


yogi-oc:
:+1:

PR #809, opened 2017-05-08
Updating pom for new Artifactory

jmcinerney:
Updated pom to point to https://openclinica.jfrog.io/openclinica


yogi-oc:
:+1: 
Good job.

PR #811, opened 2017-05-16
OC-8101 Non-alphanumeric characters not saved in repeating groups in XForms

jkeremian:
Non-alphanumeric characters not saved in repeating groups in XForms

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
+++ b/core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
@@ -436,9 +436,7 @@ public class EnketoUrlService {
 
     private String escapedValue(String value) {
         if (value != null) {
-            value = value.replaceAll("&", "&amp;");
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
```diff
@@ -533,7 +431,16 @@ private String getItemValue(ItemData itemData, CrfVersion crfVersion) {
                 itemValue = itemValue.replaceAll(",", " ");
             }
         }
-        return itemValue;
+        return escapedValue(itemValue);
+    }
+
+    private String escapedValue(String value) {
+        if (value != null) {
```
yogi-oc: concise way to do this would be
value = value.replaceAll("&", "&amp;").replaceAll(...).replaceAll()

I think it would be more readable and concise.

yogi-oc:
Good job.

PR #819, opened 2017-05-25
OC-7999 View within Record opens form directly at Query

jkeremian:
View within Record opens form directly at Query


kkrumlian:
👍

yogi-oc:
:+1: 
Good job

PR #824, opened 2017-06-01
Oc 8162 The wrong user is listed as the owner of a Query that was automatically closed when an Event CRF is deleted

jkeremian:


yogi-oc:
:+1:

PR #826, opened 2017-06-02
OC-8164 REST API call for clinical data returning 500 error

jkeremian:
REST API call for clinical data returning 500 error


yogi-oc:
:+1:

PR #827, opened 2017-06-05
OC-8188 View Within Record opens form in the same window

jkeremian:
View Within Record opens form in the same window


yogi-oc:
:+1: 
Good job.

PR #828, opened 2017-06-05
OC-8187 OCB not populating correct formLayoutOID (version) on View Event Defintion Page

jkeremian:
OCB not populating correct formLayoutOID (version) on View Event
Defintion Page


yogi-oc:
Good work.

PR #830, opened 2017-06-13
OC-8052 Oops Error after REST call to subject data after SOAP data import

jkeremian:
Oops Error after REST call to subject data after SOAP data import

### DIFF
--- a/ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
+++ b/ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
@@ -32,6 +32,7 @@ import org.akaza.openclinica.bean.submit.CRFVersionBean;
 import org.akaza.openclinica.bean.submit.DisplayItemBean;
 import org.akaza.openclinica.bean.submit.DisplayItemBeanWrapper;
 import org.akaza.openclinica.bean.submit.EventCRFBean;
+import org.akaza.openclinica.bean.submit.FormLayoutBean;
 import org.akaza.openclinica.bean.submit.ItemBean;
 import org.akaza.openclinica.bean.submit.ItemDataBean;
 import org.akaza.openclinica.bean.submit.ItemFormMetadataBean;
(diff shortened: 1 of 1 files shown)


ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
```diff
@@ -118,16 +120,16 @@ public ImportCRFDataService(DataSource ds, Locale locale) {
                 for (FormDataBean formDataBean : formDataBeans) {
 
                     CRFVersionDAO crfVersionDAO = new CRFVersionDAO(ds);
```
yogi-oc: Can we remove this if not needed?

ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
```diff
@@ -118,16 +120,16 @@ public ImportCRFDataService(DataSource ds, Locale locale) {
                 for (FormDataBean formDataBean : formDataBeans) {
 
                     CRFVersionDAO crfVersionDAO = new CRFVersionDAO(ds);
+                	FormLayoutDAO fldao= new FormLayoutDAO<>(ds);
```
yogi-oc: Do we need to create the DAO object in the loop or we can reuse it by creating it only once?

ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
jkeremian: It is need and used on line 146

ws/src/main/java/org/akaza/openclinica/web/crfdata/ImportCRFDataService.java
jkeremian: we can take bot DAOs out of the loop. I can do that.

yogi-oc:
:+1:

PR #837, opened 2017-06-21
RM1248

sjuan76:
WebServices para enrolar pacientes y crear eventos


sjuan76:
Sorry! mistake! Going to check why it did go here!

PR #842, opened 2017-07-05
Oc 8234 View Query opens Query Widget for the selected Query

jkeremian:
View Query opens Query Widget for the selected Query


kkrumlian:
Looks good

PR #854, opened 2017-07-10
dn_item_data_map column order got out of order, causing JDBC based DA…

kkrumlian:
dn_item_data_map column order got out of order, causing JDBC based DAO to fail


kkrumlian:
👍

PR #864, opened 2017-07-17
OC-8259 View Query opens Query Widget the selected Query for cascading select field

jkeremian:
View Query opens Query Widget the selected Query for cascading select
field


kkrumlian:
👍

PR #866, opened 2017-07-18
Fix OC-8262

rooney:
OC-8262 There is no Enter Data icon for a CRF in the Enter or Validate Page to begin data entry

https://jira.openclinica.com/browse/OC-8262


kkrumlian:
👍

PR #868, opened 2017-07-20
Oc 8268 Close Query from View Query page

jkeremian:
OC-7655 Form renders in View Mode.
OC-8238 View Query opens Query Widget editable but field read-only


kkrumlian:
👍

PR #869, opened 2017-07-21
Merge from master

kkrumlian:


core/src/main/java/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
```diff
@@ -199,13 +217,22 @@ public String getEditUrl(String subjectContextKey, PFormCacheSubjectContextEntry
         // Return Enketo URL
         List<FormLayoutMedia> mediaList = formLayoutMediaDao.findByEventCrfId(eventCrf.getEventCrfId());
         EnketoURLResponse eur = enketo.registerAndGetEditURL(formLayout, crfFlavor, populatedInstance, subjectContextKey, redirectUrl, markComplete, studyOid, mediaList,
-                goTo, flavor);
+                goTo, flavor, role, study, studyEvent, mode);
```
yogi-oc: Is there a way to simply this? There are just way too many parameters to this method.
Maybe we can create a RegisterEnketo object?

kkrumlian: Yeah there probably is. Let's make a note to refactor it when we are working in a single branch

yogi-oc:
:+1: 
Good work.

PR #892, opened 2017-07-31
OC-8301 View Query link for From Query opens Enketo with "View Query" as Form Title

jkeremian:
View Query link for From Query opens Enketo with "View Query" as Form
Title


yogi-oc:
:+1:

PR #912, opened 2017-08-14
- Removed is_stateful column that was not needed

yogi-oc:
- OC-8311


yogi-oc:
This would require wiping out the Bridge DB.

PR #933, opened 2017-08-23
OC-8325 Preview CRF button does not open a preview of the form

jkeremian:
Preview CRF button does not open a preview of the form


yogi-oc:
Good job

PR #938, opened 2017-08-24
OC-8333 Unable to Update query in View Query mode from Queries table when the study is Frozen

jkeremian:
Unable to Update query in View Query mode from Queries table when the
study is Frozen


kkrumlian:
👍

web/src/main/java/org/akaza/openclinica/controller/openrosa/OpenRosaSubmissionController.java
```diff
@@ -511,11 +509,10 @@ private boolean mayProceed(Study childStudy, StudySubjectBean ssBean) throws Exc
 
         if (ssBean == null) {
             logger.debug("studyStatus: " + studyStatus);
-            if (studyStatus.equalsIgnoreCase("available"))
+            if (studyStatus.equalsIgnoreCase("available") || studyStatus.equalsIgnoreCase("frozen"))
```
yogi-oc: I know it is existing code but since we're adding to it, is it possible to use the Status constants? I believe they're defined.

PR #941, opened 2017-08-24
OC-8333 Unable to Update query in View Query mode from Queries table when the study is Frozen

jkeremian:
Unable to Update query in View Query mode from Queries table when the
study is Frozen


yogi-oc:
:+1:

PR #959, opened 2017-08-30
4x more query fixes

sdibona:


web/src/main/java/org/akaza/openclinica/control/submit/ListNotesTableToolbar.java
```diff
@@ -136,7 +135,7 @@ public String enabled() {
          *      java.util.Locale)
          */
         String getIndexes() {
-            String result = "4, 5, 9, 11, 14, 16, 17, 19";
+            String result = "1, 4, 8, 10, 12, 16";
```
yogi-oc: Also need clarification here as to why this was changed.

sdibona: See comments above method.

yogi-oc:
:+1: 
Good job

PR #962, opened 2017-08-31
- Site role deletion

yogi-oc:
- Added methods to get study roles with user and study combo


sgupta2010:
kindly unsubscribe me

thanks

PR #963, opened 2017-09-01
Merging update subject / update study subject pages and updating quer…

sdibona:
…ies to work on appropriate fields.

This is not the final state of these changes, which are still in progress.  This pull request is so testers can look at the updated pages and provide feedback.

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/control/admin/UpdateSubjectServlet.java
+++ b/web/src/main/java/org/akaza/openclinica/control/admin/UpdateSubjectServlet.java
@@ -36,12 +36,12 @@ import org.apache.commons.lang.StringUtils;
 public class UpdateSubjectServlet extends SecureController {
     
     /**
-	 * 
-	 */
-	private static final long serialVersionUID = 1L;
-	// Changes
+     * 
+     */
+    private static final long serialVersionUID = 1L;
+    // Changes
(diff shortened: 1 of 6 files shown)


web/src/main/java/org/akaza/openclinica/control/admin/UpdateSubjectServlet.java
```diff
@@ -276,56 +276,56 @@ else if ( currentStudy.getStudyParameterConfig().getCollectDob().equals("2")){
...
+            if (currentStudy.getStudyParameterConfig().getSubjectPersonIdRequired().equals("required") &&
+                    !(subject.getUniqueIdentifier() == null || subject.getUniqueIdentifier().isEmpty() ) &&
+                    (uniqueIdentifier == null || uniqueIdentifier.isEmpty())) {
+                Validator.addError(errors, "uniqueIdentifier", resexception.getString("field_not_blank"));
+                
+            }
+             if (uniqueIdentifier != null && !uniqueIdentifier.isEmpty()) {
                 
+                 
+                if ( uniqueIdentifier.length() > 255){
```
yogi-oc: Can we define a constant for this (255) some where so we can change it we need to?

yogi-oc:
-- To be cleaned up later.

PR #975, opened 2017-09-06
OC-8339 Image preview not rendered in OCForm when a CRF with an uploaded image is re-opened by user

jkeremian:
Image preview not rendered in OCForm when a CRF with an uploaded image
is re-opened by user


yogi-oc:
:+1: 
Good job

PR #981, opened 2017-09-07
Adding JsonIgnoreProperties to OCUserDTO to prevent OC from breaking

kkrumlian:
Adding JsonIgnoreProperties to OCUserDTO to prevent OC from breaking in case new properties are added to SBS before being added to Bridge


kkrumlian:
👍

PR #982, opened 2017-09-11
4x update query icon colors new

sdibona:
Fixed issue where parent window query icon was not updated when a new Query was created.

Note: as with 3.x, this logic only applies to creating new queries.  Modifying an existing query does not update the parent window query icon.


kkrumlian:
👍

PR #998, opened 2017-09-18
OC-8099 Latest form version layout definition appears in the extract for each element

jkeremian:
Also, resolves the issue with loading the original version in Test when the form version is overwritten.

Latest form version layout definition appears in the extract for each
element


kkrumlian:
👍

PR #1001, opened 2017-09-19
OC-8372 Old form definition is loaded in Enketo after the version is overwritten and published

jkeremian:
Old form definition is loaded in Enketo after the version is overwritten
and published


kkrumlian:
👍

PR #1007, opened 2017-09-20
OC-8376 Form metadata file locations in data folder

jkeremian:
For this to work properly , Forms need to be republished and picture files within forms should be re uploaded to the form.
Thanks


kkrumlian:
👍

PR #1012, opened 2017-09-22
OC-8381 Clicking View Query in Queries Table at the Site-level generates Oops Error

jkeremian:
Clicking View Query in Queries Table at the Site-level generates Oops
Error


kkrumlian:
👍

PR #1027, opened 2017-09-28
- ignore studies that are not published in the change study list

yogi-oc:
- fixed the bean loading issue with liquibase


kkrumlian:
@yogi-oc Quick question this new column only pushed to tenant level study  table? or to both public and tenant ?

PR #1030, opened 2017-09-29
OC-8393 updated error messages and exluded < and > characters

jkeremian:
updated error messages and got rid of "<" and ">" characters


kkrumlian:
👍

PR #1033, opened 2017-09-29
update publish flag @ study & site level

kkrumlian:


core/src/main/java/org/akaza/openclinica/service/OdmImportServiceImpl.java
```diff
@@ -220,10 +220,28 @@ public void importOdmToOC(ODM odm, String boardId) {
         if (errors.hasErrors()) {
             throw new CustomRuntimeException("There are errors with publishing", errors);
         }
-        CoreResources.setRequestSchema("public");
+
+        //updateStudypublishedFlag(getStudyDao().findByStudyEnvUuid(study.getStudyEnvUuid()));
```
sdibona: we should get rid of the commented out code when the final code is committed.

sdibona:
👍

PR #1045, opened 2017-10-03
Run rules when importing data

kkrumlian:

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/logic/rulerunner/ImportDataRuleRunnerContainer.java
+++ b/core/src/main/java/org/akaza/openclinica/logic/rulerunner/ImportDataRuleRunnerContainer.java
@@ -1,10 +1,12 @@
 package org.akaza.openclinica.logic.rulerunner;
 
+import org.akaza.openclinica.bean.admin.CRFBean;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/akaza/openclinica/logic/rulerunner/ImportDataRuleRunnerContainer.java
```diff
@@ -1,244 +1,264 @@
...
+                seds.put(sedOid, sed);
+            }
+            ArrayList<FormDataBean> formDataBeans = studyEventDataBean.getFormData();
+            for (FormDataBean formDataBean : formDataBeans) {
+                String cvOid = formDataBean.getFormOID();
+                CRFVersionBean crfVersion;
+                if(cvs.containsKey(cvOid))
+                    crfVersion = cvs.get(cvOid);
+                else {
+                    // crfVersion = new CRFVersionDAO<String, ArrayList>(ds).findByOid(cvOid);
```
sdibona: when possible, we should remove commented out code

sdibona:
👍 
we should remove the commented out code if time allows

PR #1047, opened 2017-10-04
Manual Testing / Bug fixing - Misc Cosmetic Issues 

theywa:
- For Bridge search boxes, change the default text to be blank (instead of “Study Subject ID”). For the top left header subject search box only, make the default text be “Study Subject ID”

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/domain/xform/XformItem.java
+++ b/core/src/main/java/org/akaza/openclinica/domain/xform/XformItem.java
@@ -10,6 +10,7 @@ public class XformItem {
     String itemResponseType = null;
     String itemDataType = null;
     String readonly = null;
+    String relevant = null;
     boolean calculate = false;
     int itemOrderInForm;
     boolean required = false;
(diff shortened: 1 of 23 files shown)


theywa:
When optional fields are enabled for a study, the Add New Subject Window has a couple issues:
- The labels for the optional fields (Person ID & Date of Birth) do not match the standard fields (see below).
- If Date of Birth, Sex, or Person ID is configured to be required, those fields should have the same required style asterisk (*) as the other required fields. Make the asterisks red and put them right after the field labels.
screenshot from 2017-10-05 18-13-21

theywa:
When optional fields are enabled for a study, the Add New Subject Window has a couple issues:
- The window is not big enough to fit all of the optional fields (Person ID and Date of Birth) and the Add/Cancel buttons - some content/buttons gets pushed out of the window.  It is worse if any error messages are displayed (for example, submit the form with no data entered so that required field error messages are shown). To resolve this, make the window wide enough to fit all fields and all buttons. Make the window scale vertically so that all of the fields, buttons, and error messages fit.  Make sure to test with Person ID = Always, Sex = Always, and Date of Birth = Always and Date of Birth = Only the Year. Also test after submitting the form with missing or invalid values so that error messages appear.
screenshot from 2017-10-05 19-53-29

theywa:
Moved to https://github.com/OpenClinica/OpenClinica/pull/1059

PR #1056, opened 2017-10-05
Update rules template

kkrumlian:


core/src/main/resources/properties/rules_template.xml
```diff
@@ -1,58 +1,29 @@
...
+
 	<RuleDef OID="" Name="">
 		<Description></Description>
 		<Expression></Expression>
 	</RuleDef>
-	
-</RuleImport>
\ No newline at end of file
+
+</RuleImport>o
```
sdibona: Unexpected character at end.  May cause errors.  Should be removed.

web/src/main/resources/datainfo.properties
```diff
@@ -364,4 +364,15 @@ pformMaxSubmissionSize=100000000
 form.engine.url=http://yogi.local:8006
 ocform.adminapikey=ocrocks
 SBSUrl=http://yogi.dev-sbs.openclinica.info:8080/user-service/api/users/
```
sdibona: This URL should be updated to the production value.  They are developer specific at the moment.

sdibona:
👍

PR #1067, opened 2017-10-09
Remove versions column for View CRF Details for a CRF with multiple versions

kkrumlian:
Remove versions column for View CRF Details for a CRF with multiple versions


sdibona:
👍

PR #1080, opened 2017-10-11
OC-8418 Event-level query status carries over to other events

jkeremian:
Event-level query status carries over to other events


yogi-oc:
:+1:

PR #1084, opened 2017-10-12
Manual Testing/Bug Fixing - Misc Cosmetic Issues (Oct 12 progress)

adison93:


adison93:
The repeating event popup has some style and alignment issues. There should also be a space after the word “Occurrence” and we can remove the # character.
repeating

SDV Page (Monitors only) - the Subject search in the navigation bar is misaligned (it is not centered vertically like it is when any other page is shown).
before:
sdvbefore
after:
sdvafter

SDV Page (Monitors only) - fix favicon
before:
faviconsdv before
after:
faviconsdv before

adison93:
screenshot from 2017-10-12 16-32-12

PR #1090, opened 2017-10-13
Bug on calendar 

theywa:
The calendar icons in Add New Subject, Schedule Event, etc, are not doing anything. I click them and nothing happens
screenshot from 2017-10-13 18-21-24

I think this issue happen because duplicate element id so I try make the id more unique by adding some random number to solved this issue.


kkrumlian:
👍

PR #1097, opened 2017-10-16
Manual Testing/Bug Fixing - Misc Cosmetic Issues (Oct 16 progress)

adison93:
On View Event screen forms that are not yet started are not displaying the correct icon (see top icon for Form 9 in this screenshot). On View Subject screen, all forms (started or not started) are displaying the wrong icon. On Subject Matrix, locked event is displaying the wrong icon. The icon for Locked Forms should be the “lock” icon for all started or not started forms on View Event and View Subject. The icon for Locked Events should also be the “lock” icon on the Subject Matrix.
before:
before
after:
after


kkrumlian:
👍

PR #1098, opened 2017-10-16
Fixing issue with creating studies & hardcoded dependency on dbUser

kkrumlian:
Fixing issue with creating studies & hardcoded dependency on dbUser clinica


yogi-oc:
:+1:

PR #1101, opened 2017-10-17
OC-8420 Audit Log entries have different order when viewed by different users

jkeremian:
Audit Log entries have different order when viewed by different users


yogi-oc:
:+1: 
Good job.

PR #1103, opened 2017-10-18
Set document mode to Edge instead of IE8

theywa:
screenshot from 2017-10-18 20-41-48


theywa:
Pending after R1 clear

PR #1107, opened 2017-10-20
Manual Testing/Bug Fixing - Misc Cosmetic Issues

adison93:
View CRF Details page (in Bridge)
Remove Description field
before:
view crf details page in bridge remove description field_before
after:
view crf details page in bridge remove description field_after

Versions Table:
Remove Description and Revision Notes columns
before:
versions table remove description and revision notes columns_before
after:
screenshot from 2017-10-20 16-34-25

Items Table
Remove Description column
items before
after:
items_after

The repeating event “x2” should be made a hyperlink and placed closer to the visit status icon.
x2

Data Specialist, Data Entry Person, Investigator, Clinical Research Coordinator - These roles get the Subject Matrix on login. Can we expand the Info pane on the left bar? Without this pane expanded, the user doesn’t see the study name, status, etc. We can probably just have it open on page load all the time for all users.
ds


kkrumlian:
👍

PR #1109, opened 2017-10-23
Tasks > Monitor and Manage Data > View Study:

theywa:
- Brief Summary (in Overview and Section A) should show the study Description defined in Study Manager.
- Section B - Study Phase - Show the Display Text that is visible in the dropdown in Study Manager (example: “Phase I” not “PHASEI”)
screenshot from 2017-10-23 15-36-50

Note:
I found out that description from study-manager not saved yet to database bridge so I save it to parameter protocolDescription.(let me know if that not proper place for it)


kkrumlian:
👍

PR #1110, opened 2017-10-24
OC-8423 Cancelling import after passing import checks still updates CRF status

jkeremian:
Cancelling import after passing import checks still updates CRF status


kkrumlian:
👍

PR #1113, opened 2017-10-26
OC-8388 Queries table

theywa:
- Queries table - Add Queries to Event End Date or Start Date and Subject Enrollment Date or Date of Birth. Go to the Queries table. Filter the table by the Item Name column and type “Date”. No results are returned even though this is an exact case match for the event/subject items. Filter by “date”. This returns all of the rows for the event/subject queries even though it is not a case match for the event/subject items. To match the behavior of other columns, filters should only return results that match the case entered (so “Date” should return each of the event/subject queries and “date” should not return any of them. See OC-8388

Note: 
- Filter incasensitive work fine on all column except for "Site ID" and "Item Type" somehow it failed to show any row after rendering(I still find the row on debugging but not show any after tf.render()). This happen if using partial value filter only, work fine using full value filter.


theywa:
Queries table - Try to filter by Detailed Notes column and it gives you an Oops error page. See OC-8466

kkrumlian:
👍

PR #1115, opened 2017-10-27
Edit Manual Testing/Bug Fixing - Misc Cosmetic Issues (27/10/17 progress)

adison93:
Login as a new user. Then remove that user from all studies. Then login to Bridge as that user.  You’ll see a screen like this:
Change the text from “Your current active study is [active study], but no role.” to “Your current active study is [active study], but you have not been assigned a role.”.
before:
login as new user before
after:
login as new user after

 I'm still seeing the description column on cust7.dev-runtime. The Detailed Notes filter is working now, though. The Detailed Notes filter is working now, though.
queries table-remove description

View Subject Page
Page title is wrong. It now says “Subject Matrix for STUDYNAME”. It needs to be changed back to “View Subject: SUBJECTID”
view subject page


kkrumlian:
👍

PR #1116, opened 2017-10-27
Queries table - Problem filtering translate value

theywa:
- Queries table - Add Queries to Event End Date and Start Date. Go to the Queries table. Filter the table by the Item Name column and type “Date”. End Date and Start Date are returned. Filter by “End”. End Date is returned. Filter by “End Date” (which is an exact match for what is displayed) and nothing is returned. I expect this filter should still return all rows that match “End Date”.


kkrumlian:
👍

PR #1118, opened 2017-10-28
View CRF in preview mode

theywa:
- View CRF Details page (in Bridge)
    Versions Table: Change View icon to open Preview view
- Manage Case Report Forms page - When you click the view button for a Form version (not the first view link for the top level form), it should open the form in Preview mode like how it opens when you click Preview in Study Designer. Right now it is opening it in Read-only mode.
screenshot from 2017-10-28 19-42-15

NOTE:
There is no postMessage event fired after finish validating form in preview mode so I made it open in new tab.


kkrumlian:
👍

PR #1119, opened 2017-10-30
Edit Manual Testing/Bug Fixing - Misc Cosmetic Issues (30-31/10/17 progress)

adison93:


adison93:
The Study Subject Record section of the Study Subject View Subject Page is not displaying a subject's DoB or Enrollment Date event though those values are captured by the system.  See OC-847
before:
dob before
after:
dob after

PR #1120, opened 2017-11-01
OC Cosmetic

theywa:
Queries Table - Change order of columns when Show More is clicked - Move Date Created to between Site ID and Date Updated. Move Item Name, Item Value, and Item Type to between CRF Status and Detailed Notes.
screenshot from 2017-11-01 15-58-46

Queries Table - Change order of columns in Brief (Hide) views - Move Item Name and  Item Value to between CRF and Detailed Notes.
screenshot from 2017-11-01 15-59-19


kkrumlian:
👍

PR #1123, opened 2017-11-02
OC Cosmetic

theywa:
SDV Page - Enter a new form. Mark is Complete. It should now appear in the SDV list. Go to the SDV Page. The form is not shown. Click Clear Filter in the SDV table. Now the form appears in the list.  The form should have appeared right away. Please investigate why it is not appearing. Also, what is the Clear Filter link doing to make it appear?

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/control/submit/ListNotesTableFactory.java
+++ b/web/src/main/java/org/akaza/openclinica/control/submit/ListNotesTableFactory.java
@@ -107,8 +107,9 @@ public class ListNotesTableFactory extends AbstractTableFactory {
     protected void configureColumns(TableFacade tableFacade, Locale locale) {
 
         tableFacade.setColumnProperties("studySubject.label",
-                "discrepancyNoteBean.resolutionStatus",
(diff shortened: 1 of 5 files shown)


theywa:
- Queries Table - Please add the Discrepancy Type column back in. It was removed a while back but we need to use it again. Make the new column title for it be “Note Type”. It should be visible on the screen for Show More and for Hide. Make the Note Type column be right after Site ID column. Please move Resolution Status column to be after Note Type. Please make sure the on-screen view and print view are updated for these changes.
screenshot from 2017-11-03 17-57-53
screenshot from 2017-11-03 17-58-05

- Queries Table - Click Print icon. The columns included and their order should match the columns and order on the screen when Show More is clicked.
screenshot from 2017-11-03 17-58-29
Note:
- I found there are different translate value for "n_of_notes" on words.properties file at core and web folder so the column name show on-screen view and print view. Should I make it same? (it for column # of Notes)
- I also change the value for column # of Comments since I think it should be same of # of Notes value

kkrumlian:
👍

PR #1124, opened 2017-11-02
SDV Page fix: Once a record has been marked as SDV’d …

rooney:
… the row is missing SDV Status and action column icons (showing that it is SDV’d and allowing the user to mark as not SDV’d). Fixed.


kkrumlian:
👍

PR #1128, opened 2017-11-03
OC-8491 Bridge consumes Reason for Change submitted from Enketo

jkeremian:
Bridge consumes Reason for Change submitted from Enketo


yogi-oc:
:+1:

PR #1131, opened 2017-11-06
OC Cosmetic

theywa:
- Queries Table - Click Print icon. The columns included and their order should match the columns and order on the screen when Show More is clicked
screenshot from 2017-11-06 19-39-50


kkrumlian:
👍

PR #1135, opened 2017-11-07
OC Cosmetic

theywa:
- Queries Table - Click Print icon. The columns included and their order should match the columns and order on the screen when Show More is clicked. (Item Type column is still missing from the Print view)
screenshot from 2017-11-07 18-24-15

- Queries Table - Note Type column - We don’t need this column after all. Sorry for the confusion. Please remove it from the onscreen table and print view. However, we need to filter the data coming into the table so that only note type = “Query” is ever included. Even if the user clears filters or adds other filters, only note type = “Query” should be included in any of the views.

- The font for the values in the Resolution Status column uses small text. It doesn’t match the other columns.
screenshot from 2017-11-07 20-38-50

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/service/managestudy/ViewNotesFilterCriteria.java
+++ b/core/src/main/java/org/akaza/openclinica/service/managestudy/ViewNotesFilterCriteria.java
@@ -56,6 +56,7 @@ public class ViewNotesFilterCriteria {
         FILTER_BY_TABLE_COLUMN.put("discrepancyNoteBean.description", "description");
         FILTER_BY_TABLE_COLUMN.put("discrepancyNoteBean.detailedNotes", "detailed_notes");
         FILTER_BY_TABLE_COLUMN.put("discrepancyNoteBean.user", "user");
+        FILTER_BY_TABLE_COLUMN.put("discrepancyNoteBean.discrepancyNoteTypeId", "discrepancy_note_type_id");
     }
 
     private final Map<String, Object> filters = new HashMap<String, Object>();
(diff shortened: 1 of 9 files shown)


theywa:
- SDV Page - The View CRF column should have an icon to open the form in Review-only mode if the study/site status is Available or Frozen. It should open the form in Read-only mode if the study/site status is Design/Pending or or Locked(Please update this icon's link to open the form in View mode in the same window)
screenshot from 2017-11-08 20-17-41
Note:
I'm not sure about the view and preview mode. Maybe there should be different parameter on url but I cann't find any mode params on code servlet.

- SDV Page - The Nav Bar is broken. Notice that the “Subject Search” box is misaligned vertically and the Tasks menu can’t be closed from this page. Krikor said he thinks that previous updates were made to the non-Sprint pages, but this Spring page was not updated. Please make sure any updates to elements like this are made to both Spring and non-Spring pages.(Tasks menu still can't be closed. Mouse-over Tasks from the SDV page. Now click outside of the Tasks menu. Notice that it doesn't close. On all of the other pages, the Tasks menu closes when you do this. The same behavior should be happening when you're on the SDV page.)
Note:
The problem lies on relative path when import jquery.min.js, because there is urlPrefix on this page url(/pages) so it failed to load the jquery. Then without jquery the script would not work properly.

- The Subject ID search box is misaligned vertically on the following other pages - Edit Study Event (from View Event Page), View Subject Page, Edit Subject Details page (from View Subject Page). How come these pages do not have the same Nav Bar appears as ll of the other pages?  (Note that the Tasks menu does seem to be closeable on these pages, unlike the SDV page)

kkrumlian:
👍

PR #1139, opened 2017-11-09
OC Cosmetic

theywa:
- SDV Page - It needs to open in Enketo, but not in Preview mode like the CRF Details page. It should open with the same mode as the View link from the View Subject page for a form that is in Data Entry Started status.

- The View Subject page for Users at a site (not at the study-level) is showing the subject’s Year of Birth even though the study is configured to use Data of Birth. The Date of Birth/Year of Birth field should respect the study setting. Note that this works correctly if the subject is created at the study level rather than at a site. See OC-8507


theywa:
- We need to put the Subject Casebook section back on the View Subject Page but remove the radio button for “Printable HTML”.  The Subject Casebook section should be between the Global Subject Record section and the Go Back to Subject List link.
screenshot from 2017-11-10 19-14-31
Note: 
Also fixed icon expand and collapse when clicked

kkrumlian:
👍

PR #1140, opened 2017-11-09
Oc 8504 Display user's name for history in the current session

jkeremian:
Display user's name for history in the current session


yogi-oc:
:+1:

PR #1143, opened 2017-11-10
OC-8515     Different values between HTML and XLS audit logs present after CRF migration

jkeremian:
Different values between HTML and XLS audit logs present after CRF
migration


yogi-oc:
:+1:

PR #1146, opened 2017-11-13
Integrating form-service

kkrumlian:


core/src/main/java/org/akaza/openclinica/service/OdmImportServiceImpl.java
```diff
@@ -56,7 +54,12 @@
 import org.openclinica.ns.odm_ext_v130.v31.OCodmComplexTypeDefinitionFormLayoutRef;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.http.*;
+import org.springframework.http.converter.HttpMessageConverter;
```
yogi-oc: should we remove the next two imports since we already have http.* above imported?

yogi-oc:
:+1:

PR #1156, opened 2017-11-17
OC-8498     Study Definition is updated by failed Publish

jkeremian:
Study Definition is updated by failed Publish


kkrumlian:
👍

PR #1159, opened 2017-11-20
OC-8522 Save Queries and Reasons for Change in separate threads

jkeremian:
Save Queries and Reasons for Change in separate threads


kkrumlian:
👍

PR #1169, opened 2017-11-22
OC-Cosmetic-Master

theywa:
OC-8549 Add Subject Start Date should be conditionally required

As a data entry user, I only want to enter event date when adding a subject when I am selecting an event for the subject.
AC1: Study Event is not required in the Add Subject window.
screenshot from 2017-11-22 10-13-01

AC2: If an event is not select for Study Event, Start Date is not required.
screenshot from 2017-11-22 10-13-36

AC3: If an event is selected for Study Event, Start Date is required.
screenshot from 2017-11-22 10-14-23


kkrumlian:
👍

PR #1171, opened 2017-11-22
Edit Manual Testing/Bug Fixing - Misc Cosmetic Issues (22-24/11/17 progress)

adison93:
Home Page - Remove the number displayed after Queries assigned to me. Just display "Queries Assigned to Me" (no colon and no number) and leave it with the link to the Queries table that it currently has. See OC-8523
before:
qureistome before
after:
aueries assigned to me after

View Subject Page:
The “x2” for a repeating event with multiple occurrences is too far away from the event icon
after:
x2 before

Tasks > View Events - This page has a table for each event in the study. There is a column for Study Subject ID but it is blank for every row.
before:
before id
after:
after id


adison93:
This is not working correctly for me. What I observe is that all sites are hidden if I am currently at a Design Study or Design Site. It should be hiding only Design Sites and it should hide them regardless of the status of the Study/Site I am currently accessing.
before:
pending before
after:
pending after

web/src/main/webapp/WEB-INF/jsp/managestudy/viewStudySubject.jsp
```diff
@@ -2,58 +2,69 @@
 <%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c" %>
 <%@ taglib uri="http://java.sun.com/jsp/jstl/fmt" prefix="fmt" %>
```
kkrumlian: Like to talk about these changes before merging it in

adison93: fixed again in commit eb20d8d27658bb4c29bd8008da7fc985820ca68c

adison93:
Subject Matrix > Select an Event from the Select an Event dropdrown. Click one of the event icons from the grid. The modal that pops up has a Print link. This link needs to be removed.
after:
remove print link after

Subject Matrix > Select an Event from the Select an Event dropdrown. Click one of the CRF column filters. We need to remove the following filter options: initial data entry complete and double data entry
after:
after remove select event

Tasks > Rules > Add Rules
Remove the Review documentation for rules link.
before:
replace-remove rules before
after:
remove rules after

Tasks > Rules > Add Rules
Replace the text in the Instructions pane on the left with: “You can create rules to automatically generate patient calendars and notifications based on conditions you define. Save your rules in an XML document and upload them here.”
before:
replace-remove rules before
after:
after replace rules

adison93:
Please also remove the filter option for administrative editing.
after:
screenshot from 2017-11-24 14-28-33

core/src/main/java/org/akaza/openclinica/bean/core/DataEntryStage.java
```diff
@@ -27,7 +27,7 @@
     public static final DataEntryStage LOCKED = new DataEntryStage(7, "locked", "locked");
 
     private static final DataEntryStage[] members =
-        { UNCOMPLETED, INITIAL_DATA_ENTRY, INITIAL_DATA_ENTRY_COMPLETE, DOUBLE_DATA_ENTRY, DOUBLE_DATA_ENTRY_COMPLETE, ADMINISTRATIVE_EDITING, LOCKED };
```
kkrumlian: Can we create a new list for this. This list might be used in many places and don't want to introduce issues around this at this point

adison93: do you mean like this (commit 98d2af912253469af02ef08bc34c2054c2996c90
) sir?

kkrumlian: You got it

adison93:
Edit Study Subject Details - Error messages for Subject ID, Secondary ID, and Person ID should not display overlapping text. See OC-8571
before:
before
after:
after

kkrumlian:
👍

PR #1173, opened 2017-11-22
OC-8541 Form Update Date updated by opening form in edit mode even with no changes made

jkeremian:
Form Update Date updated by opening form in edit mode even with no
changes made


kkrumlian:
👍

PR #1176, opened 2017-11-23
OC-8541 View and Edit icons from Signature Page (Update Study Event Page) are opening the reset password page

jkeremian:
View and Edit icons from Signature Page (Update Study Event Page) are
opening the reset password page


kkrumlian:
👍

PR #1178, opened 2017-11-24
OC-8563 Failed signature kicks user out to a mostly blank page

jkeremian:
Failed signature kicks user out to a mostly blank page


kkrumlian:
👍

PR #1181, opened 2017-11-24
OC-8551 Signature page updates

jkeremian:
Signature page updated the closed query color to black.
Also added Closed-Modified column for both Subject and Event Sign pages.


yogi-oc:
:+1:

PR #1194, opened 2017-11-29
Oc 8565 Show count of queries by form and status on Signature Page

jkeremian:
Fixes for the following features:
The query count should reflect only the last entry for each query 
The Sign Subject page Edit and View links give me oops pages for a form that is not yet started. 
Sign Subject page table structure needs to be fixed.
Sign Event page issues are not as bad, but ideally will be fixed.


kkrumlian:
👍

PR #1195, opened 2017-11-29
Edit Manual Testing/Bug Fixing - Misc Cosmetic Issues (29/11/17 progress)

adison93:
The Subject ID message still overlaps if a duplicate value is entered (see screenshot I just added here). The other error messages do not appear to overlap.
before:
before overlap
after:
after overlap


adison93:
The values it is showing now are not the Subject IDs. The values I'm seeing in study PAB2 Test on cust1.dev are number 1, 2, 3, 4, 5, 6 and my subject IDs are 001, 002, 003, 004. two of my subjects have 2 events entered (6 total events for all subjects). It appears to be displaying numbering for the order the events were started. It should be displaying Subject IDs.
before:
id before
after:
after id

kkrumlian:
👍

PR #1196, opened 2017-11-29
OC-8541  Form Update Date updated by opening form using repeat_count in edit mode even with no changes made

jkeremian:
reverting     Form Update Date updated by opening form using
repeat_count in edit mode even with no changes made


yogi-oc:
:+1:

PR #1199, opened 2017-11-30
fixed view update study subject details

adison93:
fixed view update study subject details
before:
before
after:
after


yogi-oc:
:+1:

PR #1201, opened 2017-11-30
Oc 8299 Issue with repeating groups displaying data when repeat_count column is used

jkeremian:


web/src/main/java/org/akaza/openclinica/controller/openrosa/processor/FSItemProcessor.java
```diff
@@ -277,6 +281,26 @@ private void processFieldSubmissionGroupItems(ArrayList<HashMap> listOfUploadFil
                 }
             } else {
                 logger.error("Failed to lookup item: '" + itemName + "'.  Continuing with submission.");
+                int lastIndexOf = itemName.lastIndexOf(REPEATCOUNT);
```
yogi-oc: This method is getting really really big. Is it possible to break it down? At least the newly added code in a different method?

yogi-oc:
:+1:

PR #1205, opened 2017-12-06
OC-8519 Extract should contain full timestamp for queries and reasons for change

jkeremian:
Extract should contain full timestamp for queries and reasons for change


yogi-oc:
:+1:

PR #1208, opened 2017-12-13
12/12/2018 progress

adison93:


adison93:
OC-8629: When a user clicks the Add New Subject link on the Subject Matrix, the Add Subject modal window pre-populates the Start Date and Enrollment Date fields with today's date. This is different behavior from the modal window opened when a user clicks the Add Subject link in the Tasks Menu or in the top navigation bar (for Data Entry Person or Clinical Research Coordinator) and seems to be contradictory the Acceptance Criteria in OC-8549 
The Start Date and Enrollment Date fields should be blank when the Add Subject window is opened.
before:
before ad new subject
after:
after add new

adison93:
OC-8603: The Rules table needs icon cleanup. You need to upload at least one rule to see the icons in the table.
The Rule Designer icon needs to be removed completely.
The other icons need to be updated to use the current icon styles. It is a mix of old and new icons right now.
Demo video: https://drive.google.com/open?id=19KjfO1LZfsCyBENc22AT7ea4kJuMpcbP
before:
before
after:
afterrrrr rulesssssssssss

kkrumlian:
👍

PR #1210, opened 2017-12-13
OC-8627 Short and similar item names in the same form can create mismatch values in Print Crf HTML format in some cases.

jkeremian:
Short and similar item names in the same form can create mismatch values
in Print Crf HTML format in some cases.


yogi-oc:
:+1:

PR #1211, opened 2017-12-13
OC-8598 Incorrect user assignment list when executing the "Retrieve a user account" API call

jkeremian:
Incorrect user assignment list when executing the "Retrieve a user
account" API call


yogi-oc:
:+1:

PR #1236, opened 2017-12-28
OC-8573     View Within Record button is missing from the Queries Table when the Study is Locked

jkeremian:
View Within Record button is missing from the Queries Table when the
Study is Locked


yogi-oc:
:+1:

PR #1238, opened 2018-01-02
OC-8626 OCB does not populate Study Events pulldown on the Add Subject modal when opened on the Change  Study Page

theywa:
screenshot from 2018-01-02 19-17-33


kkrumlian:
👍

PR #1242, opened 2018-01-03
OC-8514 Query Widget - Assign to User List not Populating

jkeremian:
Query Widget - Assign to User List not Populating


kkrumlian:
👍

PR #1245, opened 2018-01-05
OC-8514 and OC-8622

jkeremian:
Query Widget - Assign to User List not Populating
Load external values with XPath function


yogi-oc:
:+1:

PR #1250, opened 2018-01-10
OC-8662 Fix for loading Preview without error message with clinicaldata instance

jkeremian:
Load external values with XPath function


yogi-oc:
:+1:

PR #1255, opened 2018-01-15
OC-8717: Adding duplicate site fails with generic error in Study Manager but still creates a site in Runtime

yogi-oc:
- Also increased the name limit check to 255 in the code as in the database


kkrumlian:
👍

PR #1269, opened 2018-01-19
OC-8736: Add Subject modal - start date should be required when an event is selected

adison93:
demo in: https://drive.google.com/file/d/1sKCAjgBpphvLZ8ar9pZfAbM_HZ7dPO2S/view?usp=sharing


kkrumlian:
👍

PR #1273, opened 2018-01-23
OC-8725 Reference External Data in forms using object names in the XPath

jkeremian:
Reference External Data in forms using object names in the XPath


kkrumlian:
👍

PR #1279, opened 2018-01-25
OC-8673     Can sign events that have a status of Scheduled and DE Started

jkeremian:
This code merge will fix issue with DE Started only,cause it was
broken  in OC4. With respect to Scheduled events, OC4 has the same behavior
as   OC3 today.


yogi-oc:
:+1:

PR #1288, opened 2018-01-29
OC-6322: If Study subjects enrollement date has been removed you can not extract that study subject

Taoli2018:
If Study subjects enrollement date has been removed you can not extract that study subject


kkrumlian:
👍

PR #1289, opened 2018-01-29
Extract fix, force saxon to do transformation. Change SAS transformation xslts to escape paranthesis in study oid

kkrumlian:
Extract fix, force saxon to do transformation. Change SAS transformation xslts to escape paranthesis in study oid


jkeremian:
Looks good

PR #1294, opened 2018-01-31
OC-8767 Enketo Bad Request error when trying to view archived form version

jkeremian:
Enketo Bad Request error when trying to view archived form version


yogi-oc:
:+1:

PR #1303, opened 2018-02-02
OC-AddingLoggers. To formID , formLayoutOut , formLayout obj and crf

jkeremian:
To formID , formLayoutOut , formLayout obj and crf


kkrumlian:
👍

PR #1304, opened 2018-02-04
OC-AddingLoggers.  added loggers and made modification to View Query call and Instance template

jkeremian:
added loggers and made modification to View Query call and Instance
template


yogi-oc:
:+1:

PR #1306, opened 2018-02-07
OC-8787

yogi-oc:
Logging out of Study Manager and then accessing Runtime in a new tab immediately logs the user in


svadla-oc:
👍

PR #1311, opened 2018-02-09
OC-8794

yogi-oc:
Logging in as a Monitor and then Data Specialist is not showing the correct user role in Runtime


kkrumlian:
👍

PR #1313, opened 2018-02-10
OC-8793

yogi-oc:
Issue with the result of navigating from SM to Runtime


yogi-oc:
Will introduce other sso bugs.

PR #1318, opened 2018-02-15
OC-8513, OC-8114

yogi-oc:
OC-8513
Run on Schedule not working with rules
OC-8114
Event Action or Notification Action with Run on Schedule


kkrumlian:
👍

PR #1320, opened 2018-02-23
OC-8788

yogi-oc:
New user is getting blank screen (not login screen) after setting initial password
- Also fixed stack overflow when creating a site when user role is not found in both environments


kkrumlian:
👍

PR #1321, opened 2018-02-24
OC-8840 CRC user gets error opening form with cross-form logic

jkeremian:
CRC user gets error opening form with cross-form logic


kkrumlian:
👍

PR #1322, opened 2018-02-26
OC-8821 Add new Event CRFs from the View Study Subject Overhaul page

jkeremian:
Add new Event CRFs from the View Study Subject Overhaul page


kkrumlian:
👍

PR #1325, opened 2018-02-27
OC-8817 Site ID column is only in the Detailed Queries view

jkeremian:
Site ID column is only in the Detailed Queries view


kkrumlian:
👍

PR #1326, opened 2018-02-27
OC-8620 Data Entry Person cannot Update Study Subject Details when accessing a Subject at the Study-level

theywa:
screenshot from 2018-02-27 18-27-05


kkrumlian:
👍

PR #1328, opened 2018-02-28
OC-8844 Common Events will not have Start Date

jkeremian:
Also include:
OC-8850:  Schedule Event page should not include Common Events
OC-8853:  Subject Matrix does not show Common Events
OC-8852:  Remove Common Events from View Events page


kkrumlian:
👍

PR #1334, opened 2018-03-02
OC-8814 Edge issues

theywa:
- Update favicon.ico to favicon.png
- Remove title from default returnTo callback page
- Set title each time checking user.


kkrumlian:
👍

PR #1335, opened 2018-03-02
OC-8820 Datatables tweaks (Jing's email):

rooney:
- Column highlighting on mouseover
- Per-column filtering
- Text cutoff if > 200 characters


kkrumlian:
👍

PR #1338, opened 2018-03-02
OC-8843: Welcome email notification link does not open correct Study

yogi-oc:
Welcome email notification link does not open correct Study


kkrumlian:
👍

PR #1340, opened 2018-03-06
OC-8845 & OC-8846 Oops error on Restore Event for Subject at a Site

jkeremian:
-Oops error on Restore Event for Subject at a Site
-Remove/Restore Event at Site-Level Subject hides study name from the
title bar


kkrumlian:
👍

PR #1344, opened 2018-03-09
C-8814 Edge issues(Get favicon image from URL not from local)

theywa:
screenshot from 2018-03-09 17-36-22


kkrumlian:
@theywa I would like to keep this png file within the application.

theywa:
@kkrumlian I tried use png but it's also not work. Update the code so only on edge that will used favicon from external source. Is that acceptable?

web/src/main/webapp/WEB-INF/jsp/submit/enketoFormServlet.jsp
```diff
@@ -30,6 +30,26 @@
...
+        // other browser
+        return false;
+    }
+
+    // working around for bug favicon on edge(https://jira.openclinica.com/browse/OC-8814)
+    if (isEdge()) {
+        var link = document.querySelector("link[rel*='shortcut icon']") || document.createElement('link');
+        link.type = 'image/x-icon';
+        link.rel = 'shortcut icon';
+        link.href = 'https://www.openclinica.com/wp-content/uploads/favicon192.png';
```
kkrumlian: @theywa we need to host the pngs inside the app not linked out. Can we change that.

theywa: @kkrumlian we can but it won't show up on edge. I already change all favicon.ico to favicon.png(on this commit a99bcc25da241faaa049c297bb829f363275b264) but not work. Also tried to change the type to 'image/png' instead of 'image/x-icon' but not work too. So I tried use link like on hosted auth0 login page, since it work.

kkrumlian: @theywa let's do it. I understand it won't work but I don't want us to pull the file from openclinica.com

PR #1349, opened 2018-03-12
Fix typo, 'SectionDetailsDefinition'

ccollins0601:
should be just 'SectionDetails'
relevant to https://jira.openclinica.com/browse/SUP-17241 and https://forums.openclinica.com/discussion/comment/19455

### DIFF
--- a/core/src/main/resources/properties/OpenClinica-ODM1-3-0-OC2-0-foundation.xsd
+++ b/core/src/main/resources/properties/OpenClinica-ODM1-3-0-OC2-0-foundation.xsd
@@ -182,7 +182,7 @@
 		<xs:attributeGroup ref="EventDefinitionDetailsAttributeExtension"/>
 	</xs:complexType>
 	
-	<xs:complexType name="OCodmComplexTypeDefinition-SectionDetailsDefinition">
+	<xs:complexType name="OCodmComplexTypeDefinition-SectionDetails">
 		<xs:sequence>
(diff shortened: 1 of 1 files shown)


mozz100:
Thanks for this update.  I have been trying to validate xml coming from OpenClinica's REST API against the schema, and was hitting this issue.

This modification to the .xsd is rejected by python/lxml - it renders the schema itself invalid.  Here is the error message:

```
lxml.etree.XMLSchemaParseError: Element '{http://www.w3.org/2001/XMLSchema}element', attribute 'ref': The QName value '{http://www.openclinica.org/ns/odm_ext_v130/v3.1}SectionDetails' does not resolve to a(n) element declaration., line 199
```

I think a corresponding correction is needed in line 414 (changing `SectionDetailsDefinition` to `SectionDetails`).  If I make this change locally, the schema itself is self-consistent, but my xml still doesn't validate against it.  (The next thing I hit is:

```
lxml.etree.DocumentInvalid: Element '{http://www.openclinica.org/ns/odm_ext_v130/v3.1}PresentInEventDefinition', attribute 'AllowAnonymousSubmission': The attribute 'AllowAnonymousSubmission' is not allowed., line 43
```

Please see https://gist.github.com/mozz100/020ded2d0fb562642b4c33763b0bc7b0 for the xml I am getting - it just comes from a test instance of OpenClinica for our company - the data are not real!

PR #1350, opened 2018-03-13
Study Subject details page overhaul

rooney:
- OC-8842
- OC-8820
- OC-8821


kkrumlian:
@rooney let's make sure we put  the name & number of story as part of title

PR #1355, opened 2018-03-17
OC-8923 Don't allow publish to Prod if event type was changed

jkeremian:
Don't allow publish to Prod if event type was changed


kkrumlian:
👍

PR #1369, opened 2018-03-29
OC-8729 Users get correct Runtime page on login

yogi-oc:
- Added some debug statements
- Added "return" statements where there was redirect/forward


kkrumlian:
👍

PR #1379, opened 2018-04-05
Update Delete button name on viewSubjectPage

jmcinerney:
Span within the delete button had the same name as the remove button at the event level


kkrumlian:
👍

PR #1381, opened 2018-04-09
OC-8970 Issue calling clinical data if study subject OID is set to a wildcard

jkeremian:
Issue calling clinical data if study subject OID is set to a wildcard


kkrumlian:
👍

PR #1389, opened 2018-04-10
OC-8979     Hidden forms at the site level are also hidden on the study level

jkeremian:
Hidden forms at the site level are also hidden on the study level


kkrumlian:
👍

PR #1398, opened 2018-04-12
OC-9003 issue with AddNew Button disappearing for non repeating events

jkeremian:
issue with AddNew Button disappearing for non repeating events


kkrumlian:
👍

PR #1401, opened 2018-04-17
OC-9022 Switching from AttributesMapper to ContextMapper

ksarink:
Switching from AttributesMapper to ContextMapper to add compatibility with OpenLDAP where DN (Distinguished Name) is not an attribute and cannot be accessed via the AttributesMapper

This request solves a bug where an LDAP user cannot be added due to the fact that the dn field cannot be accessed via the Spring LDAP AttributeMapper. 
screenshot_missing-dn-parameter

A nice extra feature is that one less config option is needed, since the dn field is now automatically found.


jernsting:
I had the same issue and this request fixed it. It would be great, if this would be accepted!

ksarink:
@svadla-oc May I ask why this pull request was closed?

svadla-oc:
@ksarink This was deleted automatically as part of some branch clean-up work. Could you please update the PR against the new master branch?

PR #1405, opened 2018-04-30
OC-8982 Archived Form Versions in Common Events can be edited

jkeremian:
Archived Form Versions in Common Events can be edited


kkrumlian:
👍

PR #1406, opened 2018-04-30
OC-8997 Junk characters appear in the ODM with degree symbol

jkeremian:
Junk characters appear in the ODM with degree symbol


kkrumlian:
👍

PR #1412, opened 2018-05-05
OC-9074 Publishing a study fails if taking mre than 390 seconds

jkeremian:
Publishing a study fails if taking mre than 390 seconds


svadla-oc:
👍

PR #1416, opened 2018-05-09
OC-9068 Subject Matrix is showing the same start date for multiple instances of a Repeating Event that are scheduled for different dates

jkeremian:
Subject Matrix is showing the same start date for multiple instances of a Repeating Event that are scheduled for different dates


kkrumlian:
👍

PR #1421, opened 2018-05-14
OC-9031 bugfix

rooney:
View Subject page filters should persist (AC6: All persistence is cleared when the user logs out)


yogi-oc:
:+1:

PR #1428, opened 2018-05-17
OC-8988 Include user role and username in cross-form logic clinicaldata

jkeremian:
Include user role and username in cross-form logic clinical data


yogi-oc:
:+1:

PR #1435, opened 2018-05-21
OC-9134 Persistent Filters should be cleared when the user changes to…

rooney:
… a different study environment environment


rooney:
@yogi-oc variable name refactored

yogi-oc:
:+1:

PR #1438, opened 2018-05-23
OC-9090 View Participant Details page does not load if a form has no items saved to the OC database

jkeremian:
View Participant Details page does not load if a form has no items saved to the OC database


kkrumlian:
👍

PR #1439, opened 2018-05-23
OC-9133 fix: Persistent filters are not reset if the user is logged o…

rooney:
…ut due to inactivity

https://jira.openclinica.com/browse/OC-9133


kkrumlian:
👍

PR #1445, opened 2018-05-25
OC-9134 revision: Persistent Filters should be cleared when the user …

rooney:
…changes to a different study environment environment


kkrumlian:
👍

PR #1447, opened 2018-05-25
Note:

Taoli2018:
Fixed
just FYI
after check the code change history, this issue was not caused by recent changes,I rolled bacl all rcent changes , still can reproduce this issue,
 It happened long time ago, in case other clients has same issue, then  may refere to the fix here.

### DIFF
--- a/web/src/main/webapp/includes/global_functions_javascript.js
+++ b/web/src/main/webapp/includes/global_functions_javascript.js
@@ -1,3 +1,21 @@
+  function checkDataStatus() {
+
+        objImage=document.getElementById('status_top');
+        if (objImage != null && objImage.src.indexOf('images/icon_UnsavedData.gif')>0) {
+            return confirm('<fmt:message key="you_have_unsaved_data" bundle="${resword}"/>');
+        }
+
+        return true;
+    }
+    function gotoLink() {
+
+        var OptionIndex=document.crfForm.sectionName.selectedIndex;
+        if (checkDataStatus()) {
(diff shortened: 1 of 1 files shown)


web/src/main/webapp/includes/global_functions_javascript.js
```diff
@@ -1,3 +1,21 @@
+  function checkDataStatus() {
+
+        objImage=document.getElementById('status_top');
+        if (objImage != null && objImage.src.indexOf('images/icon_UnsavedData.gif')>0) {
+            return confirm('<fmt:message key="you_have_unsaved_data" bundle="${resword}"/>');
+        }
+
+        return true;
+    }
+    function gotoLink() {
```
svadla-oc, line 10: this function is not called from anywhere. Can you look into this?

svadla-oc (changes requested):

svadla-oc:
we need to create a PR for 3.17.x and assign fix version to the ticket. Also, please add the ticket number to the commit message

web/src/main/webapp/includes/global_functions_javascript.js
Taoli2018: those 2 functions are called in many jsp pages, like initialDataEntryNw.jsp, and also can find same function definitions, but it's in html comments block

Taoli2018:
please see new PR https://github.com/OpenClinica/OpenClinica/pull/3471

PR #1448, opened 2018-05-29
OC-9169 OC4 A "Not Started" CRF can be edited in a locked event

jkeremian:
OC4 A "Not Started" CRF can be edited in a locked event


kkrumlian:
👍

PR #1449, opened 2018-05-30
OC-9164 OC4 An existing CRF in a stopped event should be editable

jkeremian:
OC4 An existing CRF in a stopped event should be editable


kkrumlian:
👍

PR #1452, opened 2018-05-31
OC-9172 Non-Repeating Common Event tables do not have paging info at …

rooney:
…the bottom

https://jira.openclinica.com/browse/OC-9172


kkrumlian:
👍

PR #1453, opened 2018-06-01
OC-8783 All data entry users should be able to restore/remove CRFs and events

jkeremian:
All data entry users should be able to restore/remove CRFs and events


kkrumlian:
👍

PR #1460, opened 2018-06-05
OC-9188 Publish error message if codes are too long for an item

jkeremian:
Publish error message if codes are too long for an item
Also changes to API calls


kkrumlian:
👍

PR #1462, opened 2018-06-06
OC 9178 Make Publish call asycnhronous , OC-9112 Run external read calculations on view mode forms

jkeremian:
OC-9112 Run external read calculations on view mode forms


yogi-oc:
:+1:

PR #1467, opened 2018-06-09
OC-9205 Export does not generate

jkeremian:
Export does not generate
This merge should resolve all the reported extract issues


kkrumlian:
👍

PR #1470, opened 2018-06-11
OC-9201 Add jini parameter to new/edit form URLs to enable jini module in form engine

jkeremian:
Add jini parameter to new/edit form URLs to enable jini module in form engine


kkrumlian:
👍

PR #1480, opened 2018-06-15
OC-9206 (feedback) - HTML Extract label remains Subject instead of Participant

rooney:
https://jira.openclinica.com/browse/OC-9206


kkrumlian:
👍

PR #1485, opened 2018-06-19
dataimport-problems

yogi-oc:
- OC-9141 Opening a Query in read-only mode should not lock the form
- Changed error from SEE_OTHER to BAD_REQUEST
- Added debug info for CRF locking


jkeremian:
Looks Good

PR #1486, opened 2018-06-19
dataimport-problems

yogi-oc:
OC-9239 New Common Events occasionally not locking for a user
- added debug and info logs. Nothing needed to be fixed as this issue was
caused by app timeout.


jkeremian:
Looks good

PR #1495, opened 2018-06-27
OC-9258 Archived Event CRF should not be able to be deleted in Runtime

jkeremian:
Archived Event CRF should not be able to be deleted in Runtime


svadla-oc:
👍

PR #1504, opened 2018-07-09
OC-8818 Editing a user profile is not reflected in query widget dropdown

yogi-oc:
OC-8818 Editing a user profile is not reflected in query widget dropdown
- Added two client properties as a default


svadla-oc:
👍

PR #1506, opened 2018-07-09
get-query-users-from-sbs

yogi-oc:
OC-8818 Editing a user profile is not reflected in query widget dropdown
- Make user service timeout configurable


svadla-oc:
👍

PR #1508, opened 2018-07-10
Oc 9318 Participant Details page based on the Participant ID method of generation

jkeremian:
Participant Details page based on the Participant ID method of generation

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/controller/StudyController.java
+++ b/web/src/main/java/org/akaza/openclinica/controller/StudyController.java
@@ -2106,12 +2106,12 @@ public class StudyController {
             logger.info("Template ID Sample :"+ wtr.toString());
         } catch (TemplateException te) {
             te.printStackTrace();
-            ErrorObj errorObject = createErrorObject("Study Object", "Template ID is not Valid\" +te.getMessage()", "templateID");
+            ErrorObj errorObject = createErrorObject("Study Object", "Template ID is not Valid" , "templateID");
             errorObjects.add(errorObject);
 
         } catch (IOException ioe) {
             ioe.printStackTrace();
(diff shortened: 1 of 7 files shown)


web/src/main/java/org/akaza/openclinica/controller/StudyController.java
```diff
@@ -657,9 +666,51 @@ private Study createSchemaStudy(HttpServletRequest request, Study study, UserAcc
         return schemaStudy;
     }
 
-    private void updateStudyConfigParameters(HttpServletRequest request, Study schemaStudy, StudyParameterConfig studyParameterConfig) {
+    private void updateStudyConfigParameters(HttpServletRequest request, Study schemaStudy, StudyParameterConfig studyParameterConfig , String templateID ,Boolean enrollmentCap) {
```
svadla-oc: I think this method should be generic such that it takes any study parameter and its key and adds an entry to the study parameter value table, if the entry doesn't exist.

svadla-oc: The way it is written currently, it doesn't scale well if we are adding new properties in the future.

web/src/main/java/org/akaza/openclinica/controller/StudyController.java
```diff
@@ -2049,6 +2093,45 @@ private Status getStatus(String myStatus) {
         return statusObj;
     }
 
+    public void verifyTemplateID(String templateID ,ArrayList<ErrorObj> errorObjects) {
+        Map<String, Object> data = new HashMap<String, Object>();
+        // Adding Sample data to validate templateID
+        data.put("siteId", "HELLO THERE");
+        data.put("siteParticipantCount", 25);
```
svadla-oc: I am wondering if we should define this map as a static map in the ParticipantIdModel object, so we have these keys defined in one place.

PR #1518, opened 2018-07-16
OC-9279 Randomly getting manifest error when opening form in edit mode

yogi-oc:
- added a debug statement to log schema name


Taoli2018:
merged

PR #1519, opened 2018-07-16
master

yogi-oc:
OC-9255 The wrong user is listed as having Locked a CRF when another user attempts to Delete a Locked CRF


Taoli2018:
merged

PR #1520, opened 2018-07-17
OC-9281 Participant audit log references participant ID as Study Subj…

rooney:
…ect ID in the audit event table

https://jira.openclinica.com/browse/OC-9281


yogi-oc:
:+1:

PR #1532, opened 2018-07-24
OC-9357 A user is able to add a Participant at a Site when the total number of Study-level Participants has reached Participant Cap

jkeremian:
A user is able to add a Participant at a Site when the total number of Study-level Participants has reached Participant Cap


svadla-oc:
👍

PR #1538, opened 2018-07-26
OC-9255 The wrong user is listed as having Locked a CRF when another user attempts to Delete a Locked CRF

yogi-oc:


Taoli2018:
merged

PR #1539, opened 2018-07-27
OC-9375 Form Permission Tags

jkeremian:
Participant Details page doesn't show Common Event Forms the user doesn't have access to


kkrumlian:
- As part of other stories let's take the caching out of the domain object
- As we do more stories like these let's look at the query and see if we can optimize

kkrumlian:
👍

PR #1540, opened 2018-07-27
OC-9385 Participant Audit Log does not include data from forms the user doesn't have access to

jkeremian:
Participant Audit Log does not include data from forms the user doesn't have access to


kkrumlian:
👍

PR #1544, opened 2018-07-31
OC-9209: Removed the alert message that was added to catch pages that…

svadla-oc:
… don't calculate enrollment cap


kkrumlian:
👍

PR #1551, opened 2018-08-02
OC-9418 User service call failure results in Runtime not allowing users to login

yogi-oc:


yogi-oc:
Putting in more comments

PR #1552, opened 2018-08-03
Merge pull request #1 from OpenClinica/master

yogi-oc:
- Get the latest
- OC-9418 User service call failure results in Runtime not allowing users to login
- Moved jwt pom to the parent pom
- Removed the duplicate StudyEnvironmentRoleDTO


yogi-oc:
Need it from another branch

PR #1553, opened 2018-08-03
OC-9418 User service call failure results in Runtime not allowing users to login

yogi-oc:
OC-9418 User service call failure results in Runtime not allowing users to login
Moved jwt pom to the parent pom
Removed the duplicate StudyEnvironmentRoleDTO


svadla-oc:
👍

PR #1555, opened 2018-08-03
OC-9385 Added TagIds to the scripts to filter out the Tagged forms for users who do not have permission tags.

jkeremian:
Added TagIds to the scripts to filter out the Tagged forms for users who do not have permission tags.


svadla-oc:
👍

PR #1556, opened 2018-08-06
OC-9290 Restore Participant page updates

theywa:
Note: On screenshot the scheduled event status name is Available(if there any available scheduled event when participant removed) but based on the code, the scheduled event status will be set to auto-removed when it's participant is removed so I used Removed status name instead of Available for any scheduled event that show up on restored page.
screenshot from 2018-08-06 14-45-11


kkrumlian:
👍

PR #1558, opened 2018-08-06
OC-9418 User service call failure results in Runtime not allowing use…

yogi-oc:
…rs to login

- Fixed the userAccount:null exception
- Fixed Auth0 login issue


kkrumlian:
👍

PR #1559, opened 2018-08-06
OC-9402 Alert box contains incorrect warning when a user tries to per…

yogi-oc:
…form a forbidden operation on a locked form in Runtime


kkrumlian:
👍

PR #1563, opened 2018-08-07
OC-9391 OC is not properly displaying the warning text in the Add Participant modal when a user tries to exceed a Participant Cap in a browser not at full size

theywa:
I see there already prevented action that would hide "Add new participant link" if already over the cap. In case  user can access the "Add new participant link", it'll look like this 
screenshot from 2018-08-07 20-47-45

Or directly access the link (/AddNewSubject), will look like this
screenshot from 2018-08-07 20-46-56


kkrumlian:
👍

PR #1575, opened 2018-08-15
Finished implementation of  major functionalities, test one simple da…

Taoli2018:
…ta import xml file successfully


kkrumlian:
👍

PR #1580, opened 2018-08-16
OC-9369     Participant Details page doesn't show Common Event Forms the user doesn't have access to

jkeremian:
Participant Details page doesn't show Common Event Forms the user doesn't have access to


kkrumlian:
👍

PR #1586, opened 2018-08-17
OC-9279 Randomly getting manifest error when opening form in edit mode

jkeremian:
Randomly getting manifest error when opening form in edit mode


kkrumlian:
👍

PR #1587, opened 2018-08-20
OC-9386 Dataset extracts do not include data from forms the user doesn't have access to

jkeremian:
Dataset extracts do not include data from forms the user doesn't have access to
This PR also includes fix for inactive action buttons on site level


kkrumlian:
👍

PR #1588, opened 2018-08-20
OC-9464: Add "No Access" page to direct users to when they don't have…

rooney:
… access to perform an action due to permission tags

https://jira.openclinica.com/browse/OC-9464


kkrumlian:
👍

PR #1589, opened 2018-08-20
OC-9388 Enter or Validate Data (aka View Event) page doesn't allow ac…

yogi-oc:
…tions on forms the user doesn't have access to


kkrumlian:
👍

PR #1593, opened 2018-08-21
OC-9484     Participant Details Page doesn't load Visit or Common Events Sections

jkeremian:
Participant Details Page doesn't load Visit or Common Events Sections


kkrumlian:
👍

PR #1600, opened 2018-08-21
OC-9479 Remove button is missing on Participant Matrix event view if permission tags are attached to a form.

yogi-oc:
OC-9479 Remove button is missing on Participant Matrix event view if permission tags are attached to a form.
OC-9474 Reassign button is inactive on participant view event page


kkrumlian:
👍

PR #1604, opened 2018-08-22
OC-9512 Two buttons missing on the Participant Matrix when sorted by Event

yogi-oc:
OC-9479 Remove button is missing on Participant Matrix event view if permission tags are attached to a form.


kkrumlian:
👍

PR #1611, opened 2018-08-23
OC-9477 & OC-9302

jmcinerney:
Reverted a change that was made to SecureController and isolated the logic to be called within the area where it was needed.


kkrumlian:
👍

PR #1620, opened 2018-08-28
OC-9536 Runtime pages load with error about external font resource

theywa:
https://jira.openclinica.com/browse/OC-9536


kkrumlian:
👍

PR #1622, opened 2018-08-28
OC-9437 Switching Participant ID setting will reflect in runtime in-session + QoL updates for QE

jmcinerney:


jmcinerney:
QoL updates add back IDs to the add participant links in the menu bar and the tasks menu. Doing so required removal of some existing javascript code that would otherwise execute.

yogi-oc:
:+1:

PR #1636, opened 2018-09-04
OC-9381, fix for updating Participant signed status if data is changed, OC-9122 event locking permissions

jmcinerney:


kkrumlian:
👍

PR #1646, opened 2018-09-07
OC-9573 Un-SDV action should not show confirmation dialog if the user…

rooney:
… doesn't have access to the form

https://jira.openclinica.com/browse/OC-9573


rooney:
@svadla-oc re-tested, still works: https://jira.openclinica.com/browse/OC-9573

PR #1648, opened 2018-09-07
OC-9569 When a Participant is signed, forms opened do not show the signature attestation

jkeremian:
When a Participant is signed, forms opened do not show the signature attestation


kkrumlian:
👍

PR #1650, opened 2018-09-10
OC-9540 more changes after testing (1.common events will be updated n…

Taoli2018:
…o matter upsert setting;2 seperate validation check from insert/update operation)


kkrumlian:
👍

PR #1657, opened 2018-09-12
OC-9617 Concurrency issue when opening a new form in edit mode

yogi-oc:


pbowen-oc:
I renamed this from OC-9049 to OC-9617. OC-9617 is the bug that was recently logged related to the previously released feature OC-9049.

PR #1664, opened 2018-09-18
Migrating fix from OC 3.14 - OC-8141: XML extract itemgroupdata issue using PGSQL 9.5

kkrumlian:
Migrating fix from OC 3.14 - OC-8141: XML extract itemgroupdata issue using PGSQL 9.5


yogi-oc:
:+1:

PR #1676, opened 2018-09-27

jkeremian:
includes fixes for the following:
At publish time , enable participate. Now, it will display participatForm property
Changed formdata status to Initial data entry when available
Updated null json object into odm object with clinicalData and subjectData (if exist) in usecases where study is not available, subject is removed , no events scheduled..


kkrumlian:
👍

PR #1680, opened 2018-10-02
OC-9646     Webservice to get scheduled participate event forms from OC for the given participant

jkeremian:
Updated the logic ApiSecurityFilter.java file for the userBean property in request session.


kkrumlian:
👍

PR #1681, opened 2018-10-02
Oc 9646

jkeremian:
changes to ApiSecurityFilter.java file to change the logic for looking into UserBean property in request session.


kkrumlian:
This PR like this won't work it's going to bring in everything.

PR #1685, opened 2018-10-03

jkeremian:
Redirected the response for edit participate forms to load the form directly.


kkrumlian:
We don't need this for now. Closing PR

PR #1686, opened 2018-10-04
OC-9646 Added Loggers to track issue with userAccount

jkeremian:
Added Loggers to track issue with userAccount


kkrumlian:
Looks good

PR #1702, opened 2018-10-22
OC-9646     Webservice to get scheduled participate event forms from OC for the given participant

jkeremian:
   Fix issue when unmarking participate forms in SBS causing site level forms in runtime not to work as expected.


kkrumlian:
👍

PR #1711, opened 2018-10-25
OC-9750 Audit Log link in the Participant Details Page does not work

rooney:
https://jira.openclinica.com/browse/OC-9750


kkrumlian:
👍

PR #1713, opened 2018-10-25
OC-9694 Site participants should see the participate forms on the dashboard only when the site is in available status

jkeremian:
Site participants should see the participate forms on the dashboard only when the site is in available status


kkrumlian:
👍

PR #1717, opened 2018-10-29
OC-9770 ODM fields for participant form property need to be changed to be called 'ParticipateForm'

jkeremian:
ODM fields for participant form property need to be changed to be called 'ParticipateForm'


kkrumlian:
👍

PR #1724, opened 2018-11-05
OC-9793 Form data submission via Participate should be logged in the audit log with participant information

jkeremian:
Form data submission via Participate should be logged in the audit log with participant information


kkrumlian:
👍

PR #1725, opened 2018-11-06
OC-9446 Removing a participant from the study should deactivate the participant's account from participate

jkeremian:
Removing a participant from the study should deactivate the participant's account from participate


kkrumlian:
👍

PR #1726, opened 2018-11-07
OC-9673 Restoring a participant into the study should reactivate the participant's account on participate

jkeremian:
Restoring a participant into the study should reactivate the participant's account on participate


kkrumlian:
👍

PR #1729, opened 2018-11-12
OC-9655 Remove Temporary Participant Form view link from Participant Details page

jkeremian:
Remove Temporary Participant Form view link from Participant Details page


pbowen-oc:
Please don't merge this now. I'd like to be able to do more testing with the Participant view for now while it is easy to access.

PR #1732, opened 2018-11-14
OC-9446 AC3: Given a removed participant, when that participant's det…

rooney:
…ails page is accessed by CRC or Investigator then participant contact information (table and the two links) should not be visible.

https://jira.openclinica.com/browse/OC-9446


kkrumlian:
👍

PR #1734, opened 2018-11-15
OC-9869 In Keycloak, can sign an event using incorrect password

yogi-oc:
OC-9867 In Keycloak, signing off on the participant record causes an oops error
- Also removed "/pages/login from AppConfig"


jmcinerney:
:+1:

PR #1737, opened 2018-11-16
OC-9868 Owner Id is not getting captured in the database when user adds a query to a field and then completes the form

jkeremian:
Owner Id is not getting captured in the database when user adds a query to a field and then completes the form


kkrumlian:
👍

PR #1739, opened 2018-11-21
OC-9853 date_completed field in event_crf table does not always get populated when an event gets marked completed

jkeremian:
date_completed field in event_crf table does not always get populated when an event gets marked completed


kkrumlian:
👍

PR #1750, opened 2018-12-05
OC-9423 Dropping not needed column in study_subject table

jkeremian:
Registering a participant with or without email address from Participant Details page


kkrumlian:
👍

PR #1752, opened 2018-12-07
OC-9885 bugfix: View query within record for a study event item query…

rooney:
… opens the incorrect record

https://jira.openclinica.com/browse/OC-9885


kkrumlian:
👍

PR #1759, opened 2018-12-13
OC-9939 Register participant users in keycloak with a static access code

jkeremian:
Register participant users in keycloak with a static access code


svadla-oc:
👍

PR #1768, opened 2018-12-19
OC-9666 Generate 6-digit access code for participant user which is unique within the customer instance

jkeremian:
Generate 6-digit access code for participant user which is unique within the customer instance


kkrumlian:
👍

PR #1775, opened 2018-12-27
OC-9977 Bugfix: Invite/Edit links doesn't work …

rooney:
…when Participate ID contains apostrophe

https://jira.openclinica.com/browse/OC-9977


rooney:
Joe/Shiva said I should fix this some other way, not by rejecting quotes in Participant ID, closing

PR #1776, opened 2018-12-27
OC-9608 Blocked calling user list in form when loading  participate flavor forms

jkeremian:
Blocked calling user list in form when loading  participate flavor forms


svadla-oc:
👍

PR #1778, opened 2018-12-28
OC-9977 bugfix: Invite/Edit links doesn't work when Participate ID co…

rooney:
…ntains apostrophe

https://jira.openclinica.com/browse/OC-9977


svadla-oc:
👍

PR #1780, opened 2019-01-02
OC-9813 Add Participant Error Workflow fix

jmcinerney:
Moved the jQuery.blockUI function call into the onload function as it was being executed before the DOM was ready.


kkrumlian:
👍

PR #1781, opened 2019-01-03
OC-9948 Forms are not available on the participate dashboard when the event status is changed from Skipped to Data Entry started

jkeremian:
Reverting changes
Also update moduleConfigDTO to return correct subdomain.


kkrumlian:
👍

PR #1782, opened 2019-01-03
OC-10001 Investigator is able to see the participant's details page for a removed participant

jkeremian:
Investigator is able to see the participant's details page for a removed participant


kkrumlian:
👍

PR #1786, opened 2019-01-04
OC-9953 bugfix: Task Menu not rendering for DS and DEP users

rooney:
https://jira.openclinica.com/browse/OC-9953


kkrumlian:
Nice work Ron I took your code and made it common cause it also existed in another place

PR #1787, opened 2019-01-04
OC-9953: Task Menu not rendering for DS and DEP users

kkrumlian:
OC-10033: New CRC taken to Change study page on first login
OC-10017: When a new configurable user role is created and assigned to a user, the role is not reflected correctly in runtime


jkeremian:
Looks good

PR #1797, opened 2019-01-10
OC-10004 Invite participant toggle should be greyed out until a valid…

rooney:
… email address or a valid phone number is provided on the modal window

https://jira.openclinica.com/browse/OC-10004


kkrumlian:
👍

PR #1807, opened 2019-01-14
OC-10055 Participant's details page breaks with a started tagged Comm…

rooney:
…on Event form

https://jira.openclinica.com/browse/OC-10055


kkrumlian:
👍

PR #1812, opened 2019-01-15
OC-10133 Import job fails with no errors in log or on UI when file co…

Taoli2018:
…ntains a participant record with no value in an optional field


kkrumlian:
👍

PR #1818, opened 2019-01-16
OC-10139 Notification Rule not firing on Participate participant

jkeremian:
Notification Rule not firing on Participate participant


kkrumlian:
👍

PR #1819, opened 2019-01-16
OC-10104:Log file should NOT be available to download until the impor…

Taoli2018:
…t has completed for all records of the file


kkrumlian:
👍

PR #1820, opened 2019-01-17
OC-10098 Hide view participant access code link from details page if …

rooney:
…participant is not registered

https://jira.openclinica.com/browse/OC-10098


kkrumlian:
👍

PR #1822, opened 2019-01-17
OC-10158 Add studyEnvUuid to the attributes sent to keycloak at the time of the participant registration

jkeremian:
Add studyEnvUuid to the attributes sent to keycloak at the time of the participant registration


kkrumlian:
👍

PR #1828, opened 2019-01-22
OC-10171:Import Silently Failing when ItemID or ItemGroupID is missing in the mapping fileMirth others integration

Taoli2018:


kkrumlian:
👍

PR #1834, opened 2019-01-24
OC-9621 Update Message displayed after user deletes Participant Form

rooney:
https://jira.openclinica.com/browse/OC-9621


svadla-oc:
👍

PR #1846, opened 2019-01-30
OC-10116 feedback (When wrong mobile phone number is entered while in…

rooney:
…viting a participant, no error message is displayed)

https://jira.openclinica.com/browse/OC-10116


kkrumlian:
👍

PR #1848, opened 2019-01-30
OC-9644 Incorrect System-Generated Participant ID is Created After participant reassignment

jkeremian:
Incorrect System-Generated Participant ID is Created After participant reassignment


kkrumlian:
👍

PR #1849, opened 2019-01-30
OC-9858 Audit add and change actions of a user on participant contact information

jkeremian:
Audit add and change actions of a user on participant contact information


svadla-oc:
👍

PR #1851, opened 2019-02-01
 OC-10269 Update handling of user role changes during a user's session

yogi-oc:
OC-10247 Study Role display missing in RT after Go from SM


kkrumlian:
👍

PR #1859, opened 2019-02-06
OC-10321 Audit events for add and change participant information in casebook and extracts

jkeremian:
Audit events for add and change participant information in casebook and extracts


kkrumlian:
👍

PR #1860, opened 2019-02-07
OC-9409 Some user roles are unable to download the Data Import template

theywa:
https://jira.openclinica.com/browse/OC-9409


kkrumlian:
👍

PR #1864, opened 2019-02-08
Performance improvement. 

jkeremian:
Performance improvement. Changed getEventCrf() fetch property from EAGER into LAZY


kkrumlian:
👍

PR #1866, opened 2019-02-11
OC-10300 Run on schedule is not working with Event action rule

jkeremian:
Run on schedule is not working with Event action rule


kkrumlian:
👍

PR #1868, opened 2019-02-11
OC-10332 Load Participant Details page metadata first to improve loading of large participant records

jkeremian:
Load Participant Details page metadata first to improve loading of large participant records


kkrumlian:
👍

PR #1871, opened 2019-02-14
OC-10343 Run on schedule is not working with Notification action rule

jkeremian:
Run on schedule is not working with Notification action rule


kkrumlian:
👍

PR #1872, opened 2019-02-15
OC-10332: Improve Participant Details Page performance

rooney:
- Fix double-entries bug reported by Krikor
- Add loading indicator


kkrumlian:
👍

PR #1873, opened 2019-02-15
OC-10366 Don't store participant details in user account table

jkeremian:
Don't store participant details in user account table


kkrumlian:
👍

PR #1882, opened 2019-02-20
OC-10254 v2 Empty Common Event created on failed import

Taoli2018:
solution option 2 ---change current implementation pattern, for common events, will finish all validation first, if pass validation, then insert into database


kkrumlian:
👍

PR #1884, opened 2019-02-20
OC-10365:XML import into non-repeating common event does not allow overwrite

Taoli2018:
XML import into non-repeating common event does not allow overwrite


kkrumlian:
👍

PR #1885, opened 2019-02-20
OC-10343 Rest call returning all participants data points are duplicating the first participant's data to the others - Present in Staging not Prod

jkeremian:
Rest call returning all participants data points are duplicating the first participant's data to the others - Present in Staging not Prod


kkrumlian:
👍

PR #1888, opened 2019-02-21
OC-10386: Oops page is shown when Importing via XML using the Runtime UI and data does not get imported

Taoli2018:


kkrumlian:
👍

PR #1904, opened 2019-03-05
OC-10441 When participant audit log is downloaded in xls format, participant's first name,email and phone number should be masked.

jkeremian:
When participant audit log is downloaded in xls format, participant's first name,email and phone number should be masked.


kkrumlian:
👍

PR #1913, opened 2019-03-12
OC-10054 SQL injection fix for ViewNotesServlet

jmcinerney:


core/src/main/java/org/akaza/openclinica/dao/managestudy/ListNotesFilter.java
```diff
@@ -44,53 +45,59 @@ public String execute(String criteria) {
     }
 
     private String buildCriteria(String criteria, String property, Object value) {
+        String filterValue = value.toString();
+
```
yogi-oc: Can you add a comment on what it strips out

yogi-oc:
:+1:

PR #1921, opened 2019-03-15
OC-10324 Participant identifier requirements for advanced search

jkeremian:
Participant identifier requirements for advanced search
Added a property in the return json object for the front end use to load the View Participant page


kkrumlian:
👍

PR #1928, opened 2019-03-19
OC-10495 API endpoint to enable/disable contacts module

jkeremian:
API endpoint to enable/disable contacts module


kkrumlian:
👍

PR #1930, opened 2019-03-20
OC-10507 Details Page Participate/Contacts table/modal not populating

jkeremian:
This PR will fix the following issues.
View Participant page does not load PII when participate is Disabled.
Change study throwing Oops error page with Root user.
/ParticipantsSearch page should not load when ContactsModule is disabled or the User does not have appropiate roles.


kkrumlian:
👍

PR #1933, opened 2019-03-21
OC-10507 Reverting changes to Secure and Keyclock Controller

jkeremian:
Reverting changes to Secure and Keyclock Controller where requestSchema was pointing to Public
when getting the status of Contacts Module


kkrumlian:
👍

PR #1942, opened 2019-03-27
OC-10227 fix: Participant Details Page General Information tables not…

rooney:
… aligned

Tables layout overhauled, rearranged, simplified.
Tested with all these 16 permutations:
- Participate is enabled/not
- Contacts is enabled/not
- It's Investigator/CRC that's logged in vs. not
- Can update Participant ID vs. not


kkrumlian:
👍

PR #1946, opened 2019-03-28
OC-10315 Advanced search:

rooney:
- fix "Show All" link not working after doing search returning empty result
- fix inputs maxlength


kkrumlian:
👍

PR #1949, opened 2019-03-29
OC-10310 Advanced search (feedback)

rooney:
- remove calls to make audit log entry (backend now does that automatically)
- move Contacts API call so it's only when Participate/Advanced Search is enabled
- move Access Code API call so it's only when View Access Code is clicked
- Underline the "Show All" link
- Clear search inputs when Show All is clicked


kkrumlian:
👍

PR #1952, opened 2019-04-01
OC-9749 fix for clinicaldata parameter

jmcinerney:


core/src/main/java/org/akaza/openclinica/service/dto/ODMFilterDTO.java
```diff
@@ -21,6 +21,8 @@ public ODMFilterDTO(String includeDns, String includeAudits, String crossFormLog
             setIncludeDN(true);
         if (includeAudits.equalsIgnoreCase("yes") || includeAudits.equalsIgnoreCase("y"))
             setIncludeAudit(true);
+        if (clinicalData.equalsIgnoreCase("no") ||  clinicalData.equalsIgnoreCase("n"))
```
yogi-oc: Do we need a clinicaData != null check here?

yogi-oc:
:+1:

PR #1955, opened 2019-04-02
OC-10388 fix Display item name in Common Events table if there is no …

rooney:
…item label

https://jira.openclinica.com/browse/OC-10388


svadla-oc:
👍

PR #1959, opened 2019-04-03
OC-10383 Confirm Participant ID from Participant Matrix(feedback)

theywa:
- changing the "Checking" button to "Continue"
- updating the window title
- adding a label for the verification text field 
- updating the error message text for when there is no match
- extra: esc key will trigger cancel button


svadla-oc:
👍

PR #1966, opened 2019-04-04
OC-10567 Item Data is not saved on form and there is no error message shown to the user

jkeremian:
Item Data is not saved on form and there is no error message shown to the user


svadla-oc:
👍

PR #1971, opened 2019-04-05
OC-10127 OC Item data table supports 3999 characters in a text field whereas Enketo form seems to fail on 2000 characters limit

jkeremian:
OC Item data table supports 3999 characters in a text field whereas Enketo form seems to fail on 2000 characters limit


svadla-oc:
👍

PR #1976, opened 2019-04-08
OC-10596 Participate status is not reflected in runtime correctly on qe dev environment

jkeremian:
OC-10596 Participate status is not reflected in runtime correctly on qe dev environment


svadla-oc:
👍

PR #1982, opened 2019-04-10
OC-10612 Incorrect errorCode by extract participant Info endpoint when studyOID or siteOID are value but user has no role define

jkeremian:
Incorrect errorCode by extract participant Info endpoint when studyOID or siteOID are value but user has no role define


svadla-oc:
👍

PR #1984, opened 2019-04-10
OC-10430 added foreign key and not null constriant to user_id column in audit_log_event table

jkeremian:
In the audit log event table, Some records have user ID as null in the database for some studies

### DIFF
--- a/core/src/main/resources/migration/tenant/4.0-2019-04-10-OC-10430.xml
+++ b/core/src/main/resources/migration/tenant/4.0-2019-04-10-OC-10430.xml
@@ -5,7 +5,16 @@
 
 
 
-	<changeSet author="jkeremian" id="4.0-2019-04-10-OC-10430-1">
+	<changeSet author="jkeremian" id="2019-04-10-OC-10430-1">
+		<sql>
+			UPDATE audit_log_event
+			SET user_id='1' WHERE user_id ='0' or user_id is Null;
+		</sql>
+	</changeSet>
+
+
+
+	<changeSet author="jkeremian" id="2019-04-10-OC-10430-2">
 		<addForeignKeyConstraint baseColumnNames="user_id"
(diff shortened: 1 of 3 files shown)


core/src/main/resources/migration/tenant/4.0-2019-04-10-OC-10430.xml
```diff
@@ -0,0 +1,27 @@
...
+								 deferrable="false" initiallyDeferred="false"
+								 referencedColumnNames="user_id"
+								 referencedTableName="user_account" referencedTableSchemaName="public"/>
+	</changeSet>
+
+
+
+
+	<changeSet author="jkeremian" id="4.0-2019-04-10-OC-10430-2">
+		<addNotNullConstraint columnName="user_id"
```
svadla-oc: this is going to fail on the servers that already have null values, right?

jkeremian: yes, it will fail for existing 0 and null values. We need to update those records in any schema that fails. thanks,Joe

core/src/main/resources/migration/tenant/4.0-2019-04-10-OC-10430.xml
```diff
@@ -0,0 +1,36 @@
...
+	xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
+
+
+
+	<changeSet author="jkeremian" id="2019-04-10-OC-10430-1">
+		<sql>
+			UPDATE audit_log_event
+			SET user_id='1' WHERE user_id ='0' or user_id is Null;
+		</sql>
+	</changeSet>
```
svadla-oc: can you move this to a separate file so it can be easily removed for production?

jmcinerney:
retest this please

svadla-oc:
👍

PR #1992, opened 2019-04-12
OC-10631 Wrong user is shown in audit log when PII entered through API, if the participant is already created

jkeremian:
Wrong user is shown in audit log when PII entered through API, if the participant is already created
This PR also includes fixes for the OC-10634 and OC-10636 in addtion to OC-10631


svadla-oc:
👍

PR #2017, opened 2019-04-22
OC 10635 Alphabets are being allowed in phone numbers when PII is entered through API. 

jkeremian:
OC 10635 : Alphabets are being allowed in phone numbers when PII is entered through API. 
OC-10637: When PII is added through API, syntax for email is not validated


svadla-oc:
👍

PR #2020, opened 2019-04-23
OC-10631 Wrong user is shown in audit log when PII entered through API, if the participant is already created

jkeremian:
Wrong user is shown in audit log when PII entered through API, if the participant is already created


svadla-oc:
👍

PR #2023, opened 2019-04-25
OC-10673 Extract participant info endpoint should be a POST method

jkeremian:
Extract participant info endpoint should be a POST method


yogi-oc:
reset this please

PR #2024, opened 2019-04-25
OC-10695 user-control extractParticipantsInfo Swagger API call by DM fails

jkeremian:
user-control extractParticipantsInfo Swagger API call by DM fails


svadla-oc:
👍

PR #2026, opened 2019-04-29
OC-10644 Invite modal changes to support SMS messaging: include the i…

rooney:
…nvite_via_sms value in the backend call payload


yogi-oc:
:+1:

PR #2027, opened 2019-04-29
OC-10674 When an invalid job uuid is passed to download file then appropriate error message should be returned

jkeremian:
When an invalid job uuid is passed to download file then appropriate error message should be returned


yogi-oc:
:+1:

PR #2028, opened 2019-04-29
OC-10455 Provide an ability to register a participant for Participate using a webservice

jkeremian:
Provide an ability to register a participant for Participate using a webservice

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/service/UserServiceImpl.java
+++ b/web/src/main/java/org/akaza/openclinica/service/UserServiceImpl.java
@@ -212,16 +212,21 @@ public class UserServiceImpl implements UserService {
             StudySubjectDetail studySubjectDetail = new StudySubjectDetail();
             studySubject.setStudySubjectDetail(studySubjectDetail);
         }
-        studySubject.getStudySubjectDetail().setFirstName(participantDTO.getFirstName());
+        if (participantDTO.getFirstName() != null)
+        studySubject.getStudySubjectDetail().setFirstName(participantDTO.getFirstName() != null ? participantDTO.getFirstName() : "");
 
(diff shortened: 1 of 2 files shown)


web/src/main/java/org/akaza/openclinica/service/UserServiceImpl.java
```diff
@@ -212,16 +212,21 @@ private StudySubject saveOrUpdateStudySubject(StudySubject studySubject, OCParti
...
-        studySubject.getStudySubjectDetail().setFirstName(participantDTO.getFirstName());
+        if (participantDTO.getFirstName() != null)
+        studySubject.getStudySubjectDetail().setFirstName(participantDTO.getFirstName() != null ? participantDTO.getFirstName() : "");
 
         if (validateService.isParticipateActive(tenantStudy)) {
-            studySubject.getStudySubjectDetail().setEmail(participantDTO.getEmail());
-            studySubject.getStudySubjectDetail().setPhone(participantDTO.getPhoneNumber());
+            if (participantDTO.getEmail() != null)
+                studySubject.getStudySubjectDetail().setEmail(participantDTO.getEmail() != null ? participantDTO.getEmail() : "");
+            if (participantDTO.getPhoneNumber() != null)
```
yogi-oc: Can you please replace != null with StringUtils.isNotEmpty?
That way it will also catch blank strings.

jkeremian: This is for API call and supposed to look for not null values and not blank values. If you have a DTO property that you don't want to overwrite it , we do not pass the property itself, if you want to overwrite or with blank value, then we pass the property with empty string.

yogi-oc:
:+1:

PR #2029, opened 2019-04-30
OC-10407 Default results count resets to 15 when user filters the list

rooney:
https://jira.openclinica.com/browse/OC-10407


jmcinerney:
:+1:

PR #2030, opened 2019-04-30
OC-10623 Provide support to send invitation message as an SMS to the …

yogi-oc:
…participant

OC-10644 Invite modal changes to support SMS messaging


jmcinerney:
:+1:

PR #2038, opened 2019-05-03
OC-10623 Provide support to send invitation message as an SMS to the …

yogi-oc:
…participant

- Added error message for future ticket


jmcinerney:
:+1:

PR #2051, opened 2019-05-09
OC-10446 Bulk participant creation API should reject the file if Part…

yogi-oc:
…icipantID column is not found or there are more than one ParticipantID columns


jkeremian:
Looks Good

PR #2060, opened 2019-05-11
OC 10802     Marking a form complete in XML import makes the event also complete when event has other forms in not started status

jkeremian:
This merge will fix OC-10802 mark complete issue, 
also 10769 noRoleSetup and noSuffPrivelege
also study event end date issues

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/controller/ImportController.java
+++ b/web/src/main/java/org/akaza/openclinica/controller/ImportController.java
@@ -87,10 +87,10 @@ public class ImportController {
     @ApiOperation( value = "To import data in an .xml file", notes = "Will import the data in a xml file " )
     @RequestMapping( value = "/clinicaldata/import", method = RequestMethod.POST )
     public ResponseEntity<Object> importDataXMLFile(HttpServletRequest request, MultipartFile file) throws Exception {
-
+        String fileNm="";
         String importXml = null;
         if (file != null) {
-            String fileNm = file.getOriginalFilename();
+             fileNm = file.getOriginalFilename();
             if (fileNm != null && fileNm.endsWith(".xml")) {
                 importXml = RestfulServiceHelper.readFileToString(file);
             } else {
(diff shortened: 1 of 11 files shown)


web/src/main/java/org/akaza/openclinica/controller/ImportController.java
```diff
@@ -130,10 +121,10 @@ public ImportController() {
 
             odmContainer = (ODMContainer) um1.unmarshal(reader);
 
-        } catch (Exception me1) {
-            me1.printStackTrace();
-            logger.info("found exception with xml transform");
-            return new ResponseEntity(ErrorConstants.ERR_INVALID_XML_FILE + "\n" + me1.getMessage(), HttpStatus.UNSUPPORTED_MEDIA_TYPE);
+        } catch (Exception e) {
+            e.printStackTrace();
```
yogi-oc: Can we remove this print statement and add logger.error("..." + e) instead of just printing the message. That will log the entire stacktrace.

web/src/main/java/org/akaza/openclinica/controller/ImportController.java
```diff
@@ -201,16 +199,22 @@ private Study getTenantStudy(String studyOid) {
...
         Study study = studyDao.findByOcOID(studyOid);
         UserAccount userAccount = userAccountDao.findById(userAccountBean.getId());
-        JobDetail jobDetail = userService.persistJobCreated(study, site, userAccount, JobType.XML_IMPORT, null);
+        JobDetail jobDetail = userService.persistJobCreated(study, site, userAccount, JobType.XML_IMPORT, fileNm);
         CompletableFuture<Object> future = CompletableFuture.supplyAsync(() -> {
-            importService.validateAndProcessDataImport(odmContainer, studyOid, siteOid, userAccountBean, schema, jobDetail);
+            try {
+                importService.validateAndProcessDataImport(odmContainer, studyOid, siteOid, userAccountBean, schema, jobDetail);
+            } catch (Exception e){
+                e.printStackTrace();
```
yogi-oc: same as above.

web/src/main/java/org/akaza/openclinica/controller/ImportController.java
jkeremian: All set , thanks

web/src/main/java/org/akaza/openclinica/controller/ImportController.java
jkeremian: All Set , thanks

PR #2082, opened 2019-05-20
OC-10801 Import should be allowed for a site participant with a study OID in the study tag

jkeremian:
Import should be allowed for a site participant with a study OID in the study tag


svadla-oc:
retest this please

PR #2083, opened 2019-05-20
OC-10762 Participant Details Page should not fail to load if there is a problem with part of the Participant's ClinicalData/Metadata

rooney:
https://jira.openclinica.com/browse/OC-10762


svadla-oc:
👍

PR #2099, opened 2019-05-24
OC-10472 OC-10426 

yogi-oc:
OC-10472 Provide an ability to capture PII for a bulk list of participants
OC-10426 Ability to register bulk list of participants for participate with or without capturing their PII


web/src/main/java/org/akaza/openclinica/service/ParticipantServiceImpl.java
```diff
@@ -377,8 +345,188 @@ public int getSubjectCount(StudyBean currentStudy) {
...
+                    p.setParticipantOid(subjectBean.getOid());
+                    p.setParticipateStatus(subjectBean.getStatus().getName());
+                    p.setStatus("Success");
+                    p.setMessage("");
+                    ocParticipateImportDTOs.add(p);
 
+                }
+                ++index;
+            }
+            writeToFile(ocParticipateImportDTOs, studyOid, fileName);
```
svadla-oc: Should we be writing to file after each participant is added such that if an uncaught exception is thrown while processing a participant the logs for previous entries are already written to file?

web/src/main/webapp/WEB-INF/keycloak.json
```diff
@@ -1,10 +0,0 @@
-{
```
svadla-oc: Is there a reason for deleting this file?

web/src/main/resources/keycloak.json
```diff
@@ -1,10 +1,10 @@
 {
```
svadla-oc: please revert this

PR #2100, opened 2019-05-27
OC-9270 Detailed note for an event level query does not get updated

theywa:
https://jira.openclinica.com/browse/OC-9270


svadla-oc:
👍

PR #2108, opened 2019-05-29
OC-10892:When a form has multiple item groups then data does not get …

Taoli2018:
…imported for a participant with no existing form data


svadla-oc:
👍

PR #2114, opened 2019-05-31
OC-10732 Ability to run a report configured in Insight from within OpenClinica(feedback)

theywa:
https://jira.openclinica.com/browse/OC-10732


web/src/main/java/org/akaza/openclinica/service/InsightReportServiceImpl.java
```diff
@@ -0,0 +1,138 @@
...
+                    fop.close();
+                }
+            } catch (IOException e) {
+                logger.error("Error closing output stream : ", e);
+            }
+        }
+    }
+
+    @Override
+    public String doPost(String api, HashMap<String, String> mapHeaders, String body, HashMap<String, String> queries) {
```
svadla-oc: this method doesn't need to be in the interface anymore since we are only calling this from within this class. You can make this method private.

svadla-oc:
👍

PR #2124, opened 2019-06-05
OC-10927 XML import code refactoring

jkeremian:
XML import code refactoring


svadla-oc:
retest this please

core/src/main/java/org/akaza/openclinica/service/crfdata/ErrorObj.java
```diff
@@ -3,17 +3,23 @@
 public class ErrorObj {
     private String code;
     private String message;
+    private String timeStamp;
```
svadla-oc: why do we need timestamp? Looks like you are not actually using it. If that's the case, could you delete this please?

svadla-oc:
👍

PR #2126, opened 2019-06-06
OC-10928 Bulk add API returning noRole setup message when user has access to the given site

yogi-oc:
OC-10929 Bulk add API returning noSufficientPrivileges error for a study level role when they are authorized users
- Fixed a bug where inivalid header will be skipped: OC-10426 Ability to register bulk list of participants for participate with or without capturing their PII


svadla-oc:
👍

PR #2127, opened 2019-06-06
OC-10926

jkeremian:
Also include fix for the following tickets:
https://jira.openclinica.com/browse/OC-10926
User can import data on a removed event CRF, with a status of invalid on Participant details page
and
https://jira.openclinica.com/browse/OC-10924  
Error reported on importing to a locked event through API is not correct
and
https://jira.openclinica.com/browse/OC-10880
Add support to validate study / site status in XML import endpoint

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/controller/ImportController.java
+++ b/web/src/main/java/org/akaza/openclinica/controller/ImportController.java
@@ -162,14 +162,14 @@ public class ImportController {
         UserAccountBean userAccountBean = utilService.getUserAccountFromRequest(request);
         ArrayList<StudyUserRoleBean> userRoles = userAccountBean.getRoles();
 
+        if (!validateService.isStudyAvailable(studyOid)) {
+            return new ResponseEntity(ErrorConstants.ERR_STUDY_OID_NOT_AVAILABLE, HttpStatus.OK);
+        }
 
         if (!validateService.isStudyOidValid(studyOid)) {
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/akaza/openclinica/controller/ImportController.java
```diff
@@ -162,14 +162,14 @@ public ImportController() {
         UserAccountBean userAccountBean = utilService.getUserAccountFromRequest(request);
         ArrayList<StudyUserRoleBean> userRoles = userAccountBean.getRoles();
 
+        if (!validateService.isStudyAvailable(studyOid)) {
+            return new ResponseEntity(ErrorConstants.ERR_STUDY_OID_NOT_AVAILABLE, HttpStatus.OK);
+        }
 
         if (!validateService.isStudyOidValid(studyOid)) {
-            return new ResponseEntity(ErrorConstants.ERR_STUDY_NOT_EXIST, HttpStatus.NOT_FOUND);
+            return new ResponseEntity(ErrorConstants.ERR_STUDY_NOT_EXIST, HttpStatus.OK);
```
svadla-oc: why are we returning 200 on error?

jkeremian: On most of the errors in endpoints we return 200. Requirements are in the ticket.
https://jira.openclinica.com/browse/OC-10880

PR #2128, opened 2019-06-06
OC 10932 If itemgroup OID is missing, no data is being added to the common event, but the common event is getting status completed

jkeremian:
If itemgroup OID is missing, no data is being added to the common event, but the common event is getting status completed

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/controller/ImportController.java
+++ b/web/src/main/java/org/akaza/openclinica/controller/ImportController.java
@@ -162,14 +162,14 @@ public class ImportController {
         UserAccountBean userAccountBean = utilService.getUserAccountFromRequest(request);
         ArrayList<StudyUserRoleBean> userRoles = userAccountBean.getRoles();
 
+        if (!validateService.isStudyAvailable(studyOid)) {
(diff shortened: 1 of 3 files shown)


web/src/main/java/org/akaza/openclinica/controller/ImportController.java
```diff
@@ -162,14 +162,14 @@ public ImportController() {
         UserAccountBean userAccountBean = utilService.getUserAccountFromRequest(request);
         ArrayList<StudyUserRoleBean> userRoles = userAccountBean.getRoles();
 
+        if (!validateService.isStudyAvailable(studyOid)) {
+            return new ResponseEntity(ErrorConstants.ERR_STUDY_OID_NOT_AVAILABLE, HttpStatus.OK);
+        }
 
         if (!validateService.isStudyOidValid(studyOid)) {
-            return new ResponseEntity(ErrorConstants.ERR_STUDY_NOT_EXIST, HttpStatus.NOT_FOUND);
+            return new ResponseEntity(ErrorConstants.ERR_STUDY_NOT_EXIST, HttpStatus.OK);
```
svadla-oc: why are we returning 200 on error?

PR #2136, opened 2019-06-07
OC-10947 Incorrect error code when repeatKey is missing in the update study event endpoint

jkeremian:
Incorrect error code when repeatKey is missing in the update study event endpoint


svadla-oc:
👍

PR #2139, opened 2019-06-08
OC - 10935 Status of event is getting completed when user asked the form status as initial data entry with Rest Import

jkeremian:
Status of event is getting completed when user asked the form status as initial data entry with Rest Import


svadla-oc:
👍

PR #2140, opened 2019-06-10
OC-10953 Participate form breaks with special characters in inputs

jkeremian:
Participate form breaks with special characters in inputs


svadla-oc:
👍

PR #2149, opened 2019-06-12
OC-10968 when formOID of visit based event is invalid or use case insensitive form id in common event, import job never completes

jkeremian:
when formOID of visit based event is invalid or use case insensitive form id in common event, import job never completes


svadla-oc:
👍

PR #2150, opened 2019-06-12
OC-10968 when formOID of visit based event is invalid or use case insensitive form id in common event, import job never completes

jkeremian:
when formOID of visit based event is invalid or use case insensitive form id in common event, import job never completes

(cherry picked from commit be3b66e888535b142e8ad1c2767e46ce45f8fe8e)


svadla-oc:
👍

PR #2151, opened 2019-06-12
OC-10975 Overwriting a non repeating common event form with different version creates a new event. 

jkeremian:
Overwriting a non repeating common event form with different version creates a new event.


svadla-oc:
👍🏻

PR #2155, opened 2019-06-13
OC-10800 Invite modal window is not present when the user first logs in

theywa:
https://jira.openclinica.com/browse/OC-10800


svadla-oc:
👍

PR #2176, opened 2019-06-25
OC-10736:Add participant endpoint should return ONLY one errorCode

Taoli2018:


jmcinerney:
jenkins retest this please

PR #2183, opened 2019-06-27
OC-10980 When study event with repeatKey 1 is sent once but contains more than one form data sections then ONLY first section should get processed

jkeremian:
When study event with repeatKey 1 is sent once but contains more than one form data sections then ONLY first section should get processed


jmcinerney:
jenkins retest this please

PR #2187, opened 2019-06-27
OC-10905 When user is invited via mobile or email then 'participant access code viewed' audit entry is also getting created

jkeremian:
When user is invited via mobile or email then 'participant access code viewed' audit entry is also getting created


jmcinerney:
jenkins retest this please

PR #2190, opened 2019-06-28
OC-10979 Remove Japan from the country list on the Invite modal window

rooney:
https://jira.openclinica.com/browse/OC-10979


svadla-oc:
👍

PR #2197, opened 2019-07-01
OC-10610:Validation of the OIDs in the skip Criteria within the mappi…

Taoli2018:
…ng file should happen before starting the import process


svadla-oc:
👍

PR #2201, opened 2019-07-02
OC-10993 User can import through REST API to a form version that is not available to a site

jkeremian:
User can import through REST API to a form version that is not available to a site


svadla-oc:
👍

PR #2203, opened 2019-07-02
OC-11072 Update Query ID display to be all numeric with no padding or prefix

jkeremian:
Update Query ID display to be all numeric with no padding or prefix


svadla-oc:
👍

PR #2211, opened 2019-07-03
OC-10717 Update the Invite/Edit Participant Window to use consistent …

rooney:
…logic for email and phone number as data collected on a form

https://jira.openclinica.com/browse/OC-10717


svadla-oc:
👍

PR #2215, opened 2019-07-05
OC-10834 OC4 - Signature not recorded when signing an event if the event status is already Signed

theywa:
https://jira.openclinica.com/browse/OC-10834


svadla-oc:
👍

PR #2216, opened 2019-07-08
OC-10915 While viewing jobs at site level, the top right hand side is appearing chopped

theywa:
https://jira.openclinica.com/browse/OC-10915


svadla-oc:
👍

PR #2232, opened 2019-07-11
OC-10992: Fixed security alerts by upgrading versions of dependencies…

svadla-oc:
… with security vulenerabilities


pom.xml
```diff
@@ -27,7 +27,7 @@
 
         <!-- Library Versions -->
         <spring.version>4.3.18.RELEASE</spring.version>
```
yogi-oc: Can we make this 4.3.24 so that spring-security and spring versions are the latest before they moved to version 5?

PR #2234, opened 2019-07-12
OC-11076 Deleting Event CRF not closing all queries threads

jkeremian:
Deleting Event CRF not closing all queries threads


svadla-oc:
👍

PR #2237, opened 2019-07-13
OC-11098 Query Manager Query Status not filtering as expected

theywa:
https://jira.openclinica.com/browse/OC-11098


svadla-oc:
👍

PR #2247, opened 2019-07-16
OC - 11119 XML Import breaking with jobs in perpetual IN_PROGRESS status

jkeremian:
XML Import breaking with jobs in perpetual IN_PROGRESS status


svadla-oc:
👍

PR #2254, opened 2019-07-18
OC -11002 Email and phone number validations in add participant and Bulk add participant endpoints to be updated

jkeremian:
Email and phone number validations in add participant and Bulk add participant endpoints to be updated


svadla-oc:
👍

PR #2261, opened 2019-07-20
OC-11099 User is able to register participant through REST API when participate module is inactive

jkeremian:
User is able to register participant through REST API when participate module is inactive


svadla-oc:
👍

PR #2263, opened 2019-07-20
OC-11168 REST API call for adding participant is not returning an error message when Bad site OID is used

jkeremian:
REST API call for adding participant is not returning an error message when Bad site OID is used


svadla-oc:
👍

PR #2264, opened 2019-07-22
OC-11163 date_completed field in event_crf table does not populate when an event is marked completed via Participate

jkeremian:
date_completed field in event_crf table does not populate when an event is marked completed via Participate


svadla-oc:
👍

PR #2265, opened 2019-07-22
OC-10991:when parameter is set as 'n' then system should not check pa…

Taoli2018:
…rticipate module status; instead should process the file and create participants with their contact info.


svadla-oc:
jenkins retest this please

svadla-oc:
👍

PR #2267, opened 2019-07-23
OC - 11187 Hide bulk add participant API from swagger UI

jkeremian:
Hide bulk add participant API from swagger UI


svadla-oc:
👍

PR #2274, opened 2019-07-25
OC-10931 Trigger Randomize for a participant only when they have valu…

yogi-oc:
…es for all stratification factors

- fixed issue with database query results order and question order not matching
- added errors for missing randomize config


svadla-oc:
👍

PR #2289, opened 2019-08-02
Oc 11162:Bulk add participant API should not proceed with processing if the given study or site in NOT available

Taoli2018:


svadla-oc:
👍

PR #2302, opened 2019-08-09
OC-10160 Status of removed participant starts with a small letter as "removed", while available participant status starts with capital letter as "Available"

theywa:
https://jira.openclinica.com/browse/OC-10160


svadla-oc:
👍

PR #2304, opened 2019-08-14
OC-3214: add implementation to support basic study wide casebook print, and fix some bugs

Taoli2018:


svadla-oc:
we have decided to not implement this for now

PR #2306, opened 2019-08-14
OC3 - business admin users can gain unauthorized access to root account

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/admin/EditUserAccountServlet.java
```diff
@@ -222,6 +226,24 @@ protected void processRequest() throws Exception {
         }
     }
 
+	/**
+	 * @param userId
+	 * @throws InsufficientPermissionException
+	 */
+	private void techAdminProtect(int userId) throws InsufficientPermissionException {
+		// only techAdmin can change root user password
+        if(userId == 1) {
```
svadla-oc: we should check for all users of type tech admin instead of user with id = 1

svadla-oc:
👍

PR #2310, opened 2019-08-15
OC-11306 Issue with Item Groups in Randomize Configurations

yogi-oc:


core/src/main/java/org/akaza/openclinica/dao/hibernate/StudyEventDao.java
```diff
@@ -149,11 +149,13 @@ public Integer findMaxOrdinalByStudySubjectStudyEventDefinition(int studySubject
 
         Query query = getCurrentSession().createQuery("select new org.akaza.openclinica.dao.hibernate.RandomizeQueryResult(s, c, ig, i) " +
                 "from StudyEvent s join s.eventCrfs c join c.formLayout.crf.itemGroups ig join c.itemDatas i " +
+                "join ig.itemGroupMetadatas igm " +
                 "where s.studyEventDefinition.oc_oid in :eventOids " +
                 "and c.studySubject.ocOid = :studySubjectOid " +
                 "and c.formLayout.crf.ocOid in :formOids " +
                 "and ig.ocOid in :itemGroups " +
                 "and i.item.ocOid in :itemOids " +
+                "and igm.item.itemId = i.item.itemId " +
```
svadla-oc: Instead of the where condition here, I think you should join itemDatas with item and then join item with itemGroupMetdatas

yogi-oc:
Not sure how that would be more efficient.

svadla-oc:
Depending on the query optimizer plan, it could potentially scan a much bigger result set and filter out rows based on your where condition.
Also, these joins would be more consistent with the other joins you already have in the query.

yogi-oc:
I am not sure about this is true for inner joins and where clause. I can
print the SQL and see if there is any difference.
https://blog.jooq.org/2019/04/09/the-difference-between-sqls-join-on-clause-and-the-where-clause/

svadla-oc:
👍

PR #2313, opened 2019-08-16
OC-11109 Audit detail for Repeating Events is not correct in XML extract

jkeremian:
Audit detail for Repeating Events is not correct in XML extract


svadla-oc:
👍

PR #2315, opened 2019-08-17
OC-11095 Adding Contact information Through Rest API on a signed participant is failing

theywa:
https://jira.openclinica.com/browse/OC-11095

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/service/ParticipantServiceImpl.java
+++ b/web/src/main/java/org/akaza/openclinica/service/ParticipantServiceImpl.java
@@ -156,8 +156,9 @@ public class ParticipantServiceImpl implements ParticipantService {
             studySubjectBean = this.getStudySubjectDao().createWithoutGroup(studySubjectBean);
 
         }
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/akaza/openclinica/service/ParticipantServiceImpl.java
```diff
@@ -156,8 +156,9 @@ public String createParticipant(SubjectTransferBean subjectTransfer,StudyBean cu
             studySubjectBean = this.getStudySubjectDao().createWithoutGroup(studySubjectBean);
 
         }
-        if(studySubjectBean!=null && !studySubjectBean.getStatus().equals(Status.AVAILABLE))
-            throw new OpenClinicaSystemException(ErrorConstants.ERR_PARTICIPANT_ID_NOT_AVAILABLE);
```
svadla-oc: instead of completely removing this check, maybe we should check for not available and not signed?

theywa: code updated

PR #2329, opened 2019-08-27
OC-11106 Login requires Logging in twice after timeout

yogi-oc:


yogi-oc:
Should be OK as is.

yogi-oc:
It's the same code with one removal in navBar.jsp which is redundant. The
formatting in GitHub is not showing things properly. I am happy to discuss.

yogi-oc:
I am OK with that but I still would like to remove that statement from
home-header.jsp which is not needed because that could reset the
crossStorage timeout to the default which is 3 seconds and not 7 seconds.
So we can create another commit for that.
Let me know.

PR #2332, opened 2019-08-28
OC-11290 Randomize configuration is loaded on Runtime startup

yogi-oc:


yogi-oc:
I was aware of that but there is existing code that is RT is using in
NotificationActionProcessor.
I agree we should move away from it going forward.

svadla-oc:
👍

PR #2338, opened 2019-08-29
OC-11304 Notification Action breaks Participate Detail page if message contains &lt; and/or &amp;

jkeremian:
Notification Action breaks Participate Detail page if message contains &lt; and/or &amp;


core/src/main/java/org/akaza/openclinica/bean/extract/odm/MetaDataReportBean.java
```diff
@@ -138,7 +138,7 @@ private String handleLoadCastor(RulesPostImportContainer rpic) {
             String newResult = result.replace("<?xml version=\"1.0\" encoding=\"UTF-8\"?>", "");
             // The result of this is being double escaped. Somewhere the Castor library is escaping the XML specific
             // characters when they have already been escaped.
-            newResult = StringEscapeUtils.unescapeXml(newResult);
```
svadla-oc: did you try removing this line completely? From the comment, it sounds like the result was being double escaped already.

jkeremian:
I'm closing this story cause anotherPR has been created https://github.com/OpenClinica/OpenClinica/pull/2345 that will replace this one.

PR #2341, opened 2019-08-30
OC 10446      Bulk participant creation API should reject the file if ParticipantID column is not found or there are more than one ParticipantID columns

jkeremian:
Bulk participant creation API should reject the file if ParticipantID column is not found or there are more than one ParticipantID columns


svadla-oc:
👍

PR #2345, opened 2019-09-04
OC-11304 Notification Action breaks Participate Detail page if message contains &lt; and/or &amp;

jkeremian:
Notification Action breaks Participate Detail page if message contains &lt; and/or &amp;


svadla-oc:
👍

PR #2349, opened 2019-09-05
OC-11400:Remove Lookup Participant API from Swagger and block at serv…

Taoli2018:
…er level


web/src/main/java/org/akaza/openclinica/controller/StudyParticipantController.java
```diff
@@ -453,7 +455,7 @@ public String startParticipantImportJob(String studyOid, String siteOid, HttpSer
 		return listStudySubjects(studyOid, siteOid, request);
 	}
 
-
+	@ApiIgnore
```
svadla-oc: use "hidden = true" argument under @ApiOperation annotation

svadla-oc:
👍

PR #2353, opened 2019-09-06
OC-11398:Error needs to be updated, when a wrong participant ID is en…

Taoli2018:
…tered to get participant information through REST API


svadla-oc:
👍

PR #2355, opened 2019-09-09
OC-11409:code refactor to use same validation logic for get bulk/spec…

Taoli2018:
…ific participant info

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/service/ValidateServiceImpl.java
+++ b/web/src/main/java/org/akaza/openclinica/service/ValidateServiceImpl.java
@@ -251,7 +251,61 @@ public class ValidateServiceImpl implements ValidateService {
         }
 
     }
+    
+    /**
+     *  this method is used when get/extract participant information
+     *  
+     * @param studyOid
+     * @param siteOid
+     * @param userAccountBean
+     * @param includeAccessCode
+     */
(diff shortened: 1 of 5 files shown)


web/src/main/java/org/akaza/openclinica/service/ValidateServiceImpl.java
```diff
@@ -251,7 +251,61 @@ public void validateStudyAndRoles(String studyOid, String siteOid, UserAccountBe
...
+    
+    /**
+     *  this method is used when get/extract participant information
+     *  
+     * @param studyOid
+     * @param siteOid
+     * @param userAccountBean
+     * @param includeAccessCode
+     */
+    public void validateStudyAndRoles(String studyOid, String siteOid, UserAccountBean userAccountBean,boolean includeAccessCode) {
```
svadla-oc: this is not just for access code, but for both contact info and access code?
maybe call it includePII?
Also, this method should only be used for reads and not writes, so call it something like validateStudyAndRolesForRead

Taoli2018: FYI
finished code change based on code review and more testing.

svadla-oc:
👍

PR #2359, opened 2019-09-11
OC-11430 Import no long firing rules when Event status is set to data entry started

jkeremian:
Import no long firing rules when Event status is set to data entry started


svadla-oc:
👍

PR #2362, opened 2019-09-13
OC-11436 Add New button is not disabled for repeating common event, w…

rooney:
…hen the participant has status of removed

https://jira.openclinica.com/browse/OC-11436


svadla-oc:
👍

PR #2366, opened 2019-09-16
OC-9834: (Update-2) Checking for existing participants Id without case sensitive while creating participants 

suthirr-openclinica:


svadla-oc:
👍

PR #2368, opened 2019-09-17
OC-9633 Using the Print option for Queries table is missing a column

rooney:
https://jira.openclinica.com/browse/OC-9633


svadla-oc:
👍

PR #2369, opened 2019-09-17
OC-10987 When user logs into the study for the fist time, common even…

rooney:
…ts did not load in participant details page

https://jira.openclinica.com/browse/OC-10987


svadla-oc:
👍

PR #2389, opened 2019-10-02
OC-11201 Matrix shows only 15 Participates where as it says 50

jkeremian:
OC-11201  Matrix shows only 15 Participates where as it says 50


svadla-oc:
👍

PR #2409, opened 2019-10-11
OC-11557     Custom columns in Queries Table should have dropdowns to filter select items

jkeremian:
Custom columns in Queries Table should have dropdowns to filter select items


svadla-oc:
👍

PR #2429, opened 2019-10-18
OC-11526 OpenClinica:Current attribute not being included in cross-form data files for Non-Repeating Visit-Based event forms

jkeremian:
OpenClinica:Current attribute not being included in cross-form data files for Non-Repeating Visit-Based event forms


svadla-oc:
👍

PR #2430, opened 2019-10-21
OC-11602: DES form marked complete on Event status change

Taoli2018:


Taoli2018:
already handled in different way after OC-11581

PR #2435, opened 2019-10-22
OC-11566 Opening form after version migration is causing duplicate entries for the form to be created in the database

jkeremian:
Opening form after version migration is causing duplicate entries for the form to be created in the database


svadla-oc:
👍

PR #2441, opened 2019-10-24
OC-11626  Reason for change added when data is imported into completed form is not appearing in Queries table consistently

jkeremian:
Reason for change added when data is imported into completed form is not appearing in Queries table consistently


svadla-oc:
👍

PR #2453, opened 2019-11-01
OC-11633 Do not display Add New button for Non-Repeating Common Event…

rooney:
… form if the form has already been entered

https://jira.openclinica.com/browse/OC-11633


svadla-oc:
👍

PR #2482, opened 2019-11-22
OC-11776:Casebook PDF Permission logic is backward

Taoli2018:


src/main/java/org/akaza/openclinica/service/ValidateServiceImpl.java
```diff
@@ -326,6 +328,20 @@ public boolean hasCRFpermissionTag(EventDefinitionCrf edc,List<String> permissio
     	
     }
     
+    public boolean hasFormAccess(EventDefinitionCrf edc,List<String> permissionTagsList) {
```
svadla-oc: Isn't there a method like this in PermissionService? Even if there isn't,  move this method to that class so that all the permission tag validation is in one place

Taoli2018: done -- move the new method to PermissionService

src/main/java/core/org/akaza/openclinica/service/PermissionServiceImpl.java
```diff
@@ -200,6 +200,20 @@ public boolean hasFormAccess(EventCrf ec, Integer formLayoutId, Integer studyEve
         return true;
     }
 
+    public boolean hasFormAccess(EventDefinitionCrf edc,List<String> permissionTagsList) {
```
svadla-oc: looks like this code is already there in the existing hasFormAccess method. Can you refactor the code such that both methods are sharing this code instead of repeating the logic?

Taoli2018: done -- both methods are sharing same common code

svadla-oc:
👍

PR #2528, opened 2019-12-11
OC-11816 The Source column in Bulk Action Logs Page is not...

rooney:
...displaying the Participant ID correctly, when there is a "-" in the Participant ID	

https://jira.openclinica.com/browse/OC-11816


svadla-oc:
👍

PR #2529, opened 2019-12-11
OC-11872:No jobs are created in Bulk actions Log page when a study le…

Taoli2018:
…vel user tries to download PDF casebook from a site.


svadla-oc:
this code may not be relevant once OC-11898 is implemented

PR #2532, opened 2019-12-12
OC-11632:Import into a Signed event un-signs the event and sets it ba…

Taoli2018:
…ck to Data Entry Started, while in UI entering data in signed event sets event status to completed


svadla-oc:
👍

PR #2536, opened 2019-12-13
OC-10523: Remove filename from job detail. Remove published jobs from…

slclinica:
… bulk action logs. Check if last in progress publish was more than 4 hours ago and stop block.


svadla-oc:
👍

PR #2543, opened 2019-12-16
OC-11814:Invalid data in XML import is setting an unscheduled Event to scheduled and the form to initial data entry

Taoli2018:


src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -381,6 +382,9 @@ public boolean validateAndProcessDataImport(ODMContainer odmContainer, String st
                                     eventCrf = updateEventCrf(eventCrf, userAccount, Status.AVAILABLE, null);
                                 }
 
+                            }else if (itemCountInForm.getInsertedUpdatedItemCountInForm() == 0 && eventCrf.getAnnotations().equals(NEW_CRF_TRACK)) {
+                            	// remove the new created blank Form
+                            	eventCrfDao.remove(eventCrf);
```
svadla-oc: Instead of creating even_crf and then deleting it from the database, try to create the event_crf only when there is a valid item data

Taoli2018: done

src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -1409,6 +1409,10 @@ private Object validateItem(ImportItemDataBean itemDataBean, CrfBean crf, EventC
                 return new DataImportReport(null, null, null, null, null, null, null, null, UPDATED, sdf_logFile.format(new Date()), null);
             }
         } else {
+        	// only created new event crf once
+        	if(eventCrf.getEventCrfId() == 0) {
+        		eventCrf = eventCrfDao.saveOrUpdate(eventCrf);	
+        	}
```
svadla-oc: move this logic to createItemData method so it will be applied to all places that call this method

Taoli2018: done

svadla-oc:
👍

PR #2546, opened 2019-12-17
OC-11920: Saving some of the default values in the db is removed

suthirr-openclinica:
Saving some of the default values in the db is removed and Some places creating redundant child StudyParameterValues is removed


svadla-oc:
👍

PR #2548, opened 2019-12-18
OC-9676 Allow change of Participant ID while auto generated is selected

theywa:
https://jira.openclinica.com/browse/OC-9676


svadla-oc:
👍

PR #2554, opened 2019-12-19
OC-11780:User able to import data into removed event

Taoli2018:


src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -1017,6 +1017,8 @@ private ErrorObj checkEventAvailable(StudyEvent studyEvent) {
         ErrorObj errorObj = null;
 
         if (studyEvent != null && (
+        		// OC-11780, for visit and just scheduled event(before enter any data),UI side will only update status of StudyEvent,because no CRF yet 
+        		studyEvent.getStatusId()==Status.DELETED.getCode() ||
```
svadla-oc: looks like this issue was created by a change for OC-10918 https://github.com/OpenClinica/OpenClinica/commit/d9b4db7b66b013622306245169ec07be4398a3e3#diff-754be736d631684cb880a2cbec793860. Please make sure that this new change doesn't break OC-10918

Taoli2018: yes, I tested OC-10918 complete form still working, because for both complete and removed event, the event status  all meet the logic check: studyEvent.getStatusId() != Status.AVAILABLE.getCode(), so we need to remove the general logic check to make the complete form  import work, and added more specific removed status logic to STOP the removed event.

svadla-oc:
👍

PR #2555, opened 2019-12-19
OC-11929 After failed publish, the customer is blocked to re-publish due to technical reason

jkeremian:
After failed publish, the customer is blocked to re-publish due to technical reason


src/main/java/core/org/akaza/openclinica/service/crfdata/EnketoUrlService.java
```diff
@@ -205,6 +205,7 @@ public FormUrlObject getActionUrl(String subjectContextKey, PFormCacheSubjectCon
         if(eventCrf==null){
             UserAccount userAccount = userAccountDao.findByUserId(ub.getId());
             eventCrf= createEventCrf(formLayout,studyEvent,subject,userAccount);
+            logger.info("creating new event crf {}",eventCrf.getEventCrfId());
```
svadla-oc: I think we should print event and crf id too, so we know why they don't match up.

svadla-oc:
I am merging this for now. Let's add more logging separately

PR #2563, opened 2019-12-23
OC-11871 Support Participant IDs using random numbers

jkeremian:
Support Participant IDs using random numbers

### DIFF
--- a/src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
+++ b/src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
@@ -1,0 +1,36 @@
+package org.akaza.openclinica.controller.helper;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+import java.util.Random;
+
+@JsonIgnoreProperties
+public class TemplateHelper {
+
+
+    public Integer random(int digit_number) {
+        switch (digit_number) {
+            case 1:
+                return new Random().nextInt(9);
+            case 2:
+                return new Random().nextInt(99);
+            case 3:
+                return new Random().nextInt(999);
+            case 4:
+                return new Random().nextInt(9999);
+            case 5:
+                return new Random().nextInt(99999);
+            case 6:
+                return new Random().nextInt(999999);
+            case 7:
+                return new Random().nextInt(9999999);
+            case 8:
+                return new Random().nextInt(99999999);
+            case 9:
+                return new Random().nextInt(999999999);
+            default:
+                return null;
+        }
+    }
+
+}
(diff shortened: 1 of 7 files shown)


src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Random;
+
+@JsonIgnoreProperties
+public class TemplateHelper {
+
+
+    public Integer random(int digit_number) {
+        switch (digit_number) {
+            case 1:
+                return new Random().nextInt(9);
```
svadla-oc: apparently it is better to use ThreadLocalRandom.current().nextInt(min, max + 1).
see: https://stackoverflow.com/questions/363681/how-do-i-generate-random-integers-within-a-specific-range-in-java

src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
```diff
@@ -0,0 +1,36 @@
...
+            case 2:
+                return new Random().nextInt(99);
```
svadla-oc: Also the value of the argument is exclusive, so this code will never generate 99 as the random number. You should set this to 100.

src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
```diff
@@ -0,0 +1,36 @@
...
+            case 3:
+                return new Random().nextInt(999);
```
svadla-oc: Instead of all these case statements, maybe write some code to find the max value based on number of digits. Maybe something like: Integer.parseInt(StringUtils.repeat("9", digit_number));

src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
```diff
@@ -0,0 +1,36 @@
...
+            case 6:
+                return new Random().nextInt(999999);
+            case 7:
+                return new Random().nextInt(9999999);
+            case 8:
+                return new Random().nextInt(99999999);
+            case 9:
+                return new Random().nextInt(999999999);
+            default:
+                return null;
```
svadla-oc: You may want to throw an error instead of returning null as this could result in a valid template with value "null" for random number.

src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
jkeremian: I updated the code based on your feedback and it worked great and looks much shorter , cleaner and professional. Thank you for your feedback , it was very helpful.

src/main/java/org/akaza/openclinica/controller/helper/TemplateHelper.java
jkeremian: same comment as before

svadla-oc:
👍

PR #2565, opened 2019-12-27
OC-11961 Refactor casebook pdf print UI side code to call correct end…

rooney:
…points

https://jira.openclinica.com/browse/OC-11961


svadla-oc:
👍

PR #2570, opened 2019-12-31
OC-11580 Auto-close queries when a form is migrated to a version that does not contain the item that is queried

theywa:
[OC-11580](https://jira.openclinica.com/browse/OC-11580)


svadla-oc:
👍

PR #2573, opened 2020-01-02
OC-11990 Query Table column that displays a numeric value is not sorting

jkeremian:
Query Table column that displays a numeric value is not sorting


svadla-oc:
👍

PR #2577, opened 2020-01-03
Fix jQuery errors

rooney:
jQuery errors on Participant Matrix:
- when switching between single-event and all-events view
- on the navbar (might affect other pages too)


svadla-oc:
👍

PR #2582, opened 2020-01-03
OC-11814:Invalid data in XML import is setting an unscheduled Event to scheduled and the form to initial data entry

Taoli2018:


svadla-oc:
👍

PR #2583, opened 2020-01-03
OC-11757: StudySubjects are fetched through hibernate instead of Jdbc call for making the timestamp work correctly

suthirr-openclinica:


svadla-oc:
👍

PR #2594, opened 2020-01-08
OC-10066 Only send form data that is needed in the external form data file sent to Enketo

jkeremian:
Only send form data that is needed in the external form data file sent to Enketo


src/main/java/core/org/akaza/openclinica/domain/xform/dto/Head.java
```diff
@@ -4,6 +4,7 @@
 
     private String title;
     private Model model;
+    private String crossform_reference;
```
svadla-oc: looks like it is called crossform_references (plural) in form-service. See https://github.com/OpenClinica/form-service/pull/82/files

svadla-oc:
👍

PR #2602, opened 2020-01-10
OC-11958: The RuleSetBean Hibernate instance got updated and it was auto-flushed and causing error is now resolved

suthirr-openclinica:


svadla-oc:
👍

PR #2606, opened 2020-01-13
OC-11934: Lazy initialization exception is removed and Item based scheduledRule processing is temporarily removed

suthirr-openclinica:


svadla-oc:
👍

PR #2629, opened 2020-01-20
OC-12139: Fix redirect link for study monitors when permission get re…

slclinica:
…voked while they are using checkbox to SDV forms


svadla-oc:
👍

PR #2631, opened 2020-01-20
OC-12135 When study has two forms in non repeating common event, and user imports data through REST API through xml having data for both forms, job is indefinitely stuck in INPROGRESS

jkeremian:
When study has two forms in non repeating common event, and user imports data through REST API through xml having data for both forms, job is indefinitely stuck in INPROGRESS


svadla-oc:
👍

PR #2635, opened 2020-01-20
OC-12032 Manual Data Import importing data and marking form complete when the item is missing 4digit random number

jkeremian:
Manual Data Import importing data and marking form complete when the item is missing 4digit random number


svadla-oc:
👍

PR #2668, opened 2020-02-03
OC-12203 API Endpoint ans Casebook calling for permission tagged data from site level not returning information

jkeremian:
API Endpoint ans Casebook calling for permission tagged data from site level not returning information


jmcinerney:
jenkins retest this please

PR #2687, opened 2020-02-06
OC-11584: Filtering itemName is altered to add provision for filtering by studyEvent itemname like startdate or enddate

suthirr-openclinica:


svadla-oc:
👍

PR #2713, opened 2020-02-13
OC-12275 Locked site makes Form review mode instead of Read-Only

jkeremian:
Locked site makes Form review mode instead of Read-Only


svadla-oc:
👍

PR #2716, opened 2020-02-18
OC-10267 Delete action should NOT be available to a form in locked status

jkeremian:
Delete action should NOT be available to a form in locked status


svadla-oc:
👍

PR #2731, opened 2020-02-25
OC-12261: Show reassign button only when more than one version or whe…

slclinica:
…n other versions of crf is available. Fix reassign icon for common events.


svadla-oc:
👍

PR #2736, opened 2020-02-27
OC-12332 Study level user should not be able to enter data in Locked Site

jkeremian:
Study level user should not be able to enter data in Locked Site


svadla-oc:
👍

PR #2740, opened 2020-02-28
Merging OC-12332 and OC-12338 into 4.10.3

jmcinerney:
Study level user should not be able to enter data in Locked Site


svadla-oc:
👍

PR #2745, opened 2020-03-02
OC-11023

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectsServlet.java
```diff
@@ -137,7 +137,26 @@ private void createTable() {
...
+        /**
+         * http://localhost:8080/OpenClinica/ListStudySubjects?module=admin&maxRows=25&showMoreLink=true&findSubjects_tr_=true&findSubjects_p_=1&findSubjects_mr_=25
+         */
+       
+        
+        // update session table attributes     
+        HashMap subjectTableSettings = getSubjectTableSettings();
+        
+        String findSubjectsHtml = null;
+        OCRequestWrapper requestWrapper = null;
```
svadla-oc, line 150: why do you need a custom request wrapper?

web/src/main/java/org/akaza/openclinica/control/submit/OCRequestWrapper.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+
+/***
+ * Request wrapper enabling the update of a request-parameter. 
+ *
+ */
+final class OCRequestWrapper
```
svadla-oc, line 13: why do you need to update request parameters?

Taoli2018: because I need keep the previous table configurations and pass to next UI page, and for the 3rd party lib:
new TableFacadeImpl(getTableName(), request);
It was passed in by request parameters

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectsServlet.java
Taoli2018: you can't  directly  intercept live request to modify/update request parameters, will  throw   errors, but use custom request wrapper can achieve this

web/src/main/java/org/akaza/openclinica/control/submit/OCRequestWrapper.java
svadla-oc: It looks like you can set the attributes on TableFacadeImpl directly. You could fetch the settings from the session and set them directly on that object in ListStudySubjectTableFactory.java. Look at SDVUtil for example

Taoli2018: Here I have different implementation, new PR:
https://github.com/OpenClinica/OpenClinica/pull/2885, if go with new PR, we can close this PR

Taoli2018:
closed old PR

PR #2747, opened 2020-03-06
OC-12272 Update SDV table filters to not push columns to be too wide

rooney:
https://jira.openclinica.com/browse/OC-12272


svadla-oc:
👍

PR #2752, opened 2020-03-11
fix OC-12390: SDV - Detailed View item table not displaying correct r…

rooney:
…epeating occurrence data for participant

https://jira.openclinica.com/browse/OC-12390


svadla-oc:
👍

PR #2757, opened 2020-03-16
OC-12480 site monitor can not log in

jkeremian:
OC-12480 site monitor can not log in
OC-12479 Common event is not loading and user can not enter data in visit based events


svadla-oc:
👍

PR #2758, opened 2020-03-16
Oc 12143

slclinica:


src/main/java/org/akaza/openclinica/service/ValidateServiceImpl.java
```diff
@@ -429,6 +434,44 @@ public void validateStudyAndRolesForRead(String studyOid, String siteOid, UserAc
...
+        }
+        if (!isSiteOidValidSiteLevelOid(siteOid)) {
+            throw new OpenClinicaSystemException(ErrorConstants.ERR_SITE_NOT_Valid_OID);
+        }
+        if (!isStudyToSiteRelationValid(studyOid, siteOid)) {
+            throw new OpenClinicaSystemException(ErrorConstants.ERR_STUDY_TO_SITE_NOT_Valid_OID);
+        }
+        if (!isUserHasAccessToStudy(userRoles, studyOid) && !isUserHasAccessToSite(userRoles, siteOid)) {
+            throw new OpenClinicaSystemException(ErrorConstants.ERR_NO_ROLE_SETUP);
+        } else if (!isUserHas_CRC_INV_DM_DEP_DS_RoleInSite(userRoles, siteOid)) {
```
svadla-oc: why did you create a new method for validating roles for extractParticipantInfo? Looks like you have changed this else-if condition for the existing read method to not check for monitor role. Wouldn't this break the existing code?

slclinica: validateStudyAndRolesForRead originally didn't allow study and site monitors to run the API, I changed it to allow site and study monitors to have access. But then I realized validateStudyAndRolesForRead is also called by the StudyParticipantController which we do not want to give monitors access, so I made a new method for it and that new method gets called by UserController.

svadla-oc: got it

svadla-oc:
👍

PR #2769, opened 2020-03-22
OC-12513 when user enters data in common event, then common event status stays not scheduled

jkeremian:
when user enters data in common event, then common event status stays not scheduled


jkeremian:
Closing this PR cause it was merged off of another PR

PR #2770, opened 2020-03-22
OC-12511 Event Status: In participant matrix, event status is not reflected.

jkeremian:
Event Status: In participant matrix, event status is not reflected.


jkeremian:
Closing this PR cause it was already merged part of another PR

PR #2771, opened 2020-03-22
OC-12514 when user schedules event through API no report is generated in bulk action log page, and status of job is Failed

jkeremian:
OC-12514 when user schedules event through API no report is generated in bulk action log page, and status of job is Failed
OC-12515 manual import not working on qe -dev env


svadla-oc:
👍

PR #2773, opened 2020-03-23
OC-12272 AC3: When the user mouses over the filter that is truncated...

rooney:
OC-12272 AC3: When the user mouses over the filter that is truncated, the full value will be displayed as hover text.

https://jira.openclinica.com/browse/OC-12272


svadla-oc:
👍

PR #2774, opened 2020-03-23
OC-11971 SDV Detailed View item table (AC12)

rooney:
https://jira.openclinica.com/browse/OC-11971

AC12: When displaying Brief Description for an item (see AC4), if the Item's Brief Description is null, the display should be "<item Label> (<Item Name>) <ordinal>". If the Item's label is also null, then the display should be "(<Item Name>) <ordinal>".


svadla-oc:
👍

PR #2786, opened 2020-03-27
OC-12549 Deleting Event CRF data renders blocking new data entry

jkeremian:
Deleting Event CRF data renders blocking new data entry


svadla-oc:
👍

PR #2788, opened 2020-03-27
OC-12546 audit log status case is different on staging and dev

jkeremian:
audit log status case is different on staging and dev


svadla-oc:
👍

PR #2793, opened 2020-03-30
OC-12567 oops error is reported while viewing form in participant matrix, with event status not scheduled

jkeremian:
oops error is reported while viewing form in participant matrix, with event status not scheduled


svadla-oc:
👍

PR #2794, opened 2020-03-30
OC-12568 User is unable to do version change for CRF

jkeremian:
User is unable to do version change for CRF

### DIFF
--- a/src/main/java/core/org/akaza/openclinica/service/StudyEventServiceImpl.java
+++ b/src/main/java/core/org/akaza/openclinica/service/StudyEventServiceImpl.java
@@ -943,6 +943,51 @@ public class StudyEventServiceImpl implements StudyEventService {
 
     }
 
+public void convertStudyEventStatus(String value, StudyEvent studyEvent){
+        StudyEventWorkflowStatusEnum workflowStatus=convertOriginalStudyEventStatusToWorkflowSatus(value);
+        if (workflowStatus!=null) {
+            studyEvent.setWorkflowStatus(workflowStatus);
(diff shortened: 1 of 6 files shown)


src/main/java/core/org/akaza/openclinica/service/StudyEventServiceImpl.java
```diff
@@ -943,6 +943,51 @@ public void scheduleOrUpdateBulkEvent(MultipartFile file, Study study, String si
...
+        }else{
+            if (value.equals("7")) {
+                studyEventBean.setLocked(Boolean.TRUE);
+            }else if (value.equals("8")){
+                studyEventBean.setSigned(Boolean.TRUE);
+            }
+        }
+    }
+
+    public StudyEventWorkflowStatusEnum convertOriginalStudyEventStatusToWorkflowSatus(String value) {
```
svadla-oc: Instead of doing this conversion, is it possible to migrate the audit log data as well?

jkeremian: I thought about migrating audit log but it will not be straight forward , since we are dividing study event 1 status type into 5 different status types. (Workflow, removed, archived, locked, signed). Migration time, we need to insert new records in auditlog table for each new status type. The same issue is with EventCrf statuses.

svadla-oc: Got it.

svadla-oc:
👍

PR #2796, opened 2020-03-30
OC-12560 Status appearing all Caps in Participant Matrix

jkeremian:
Status appearing all Caps in Participant Matrix


svadla-oc:
👍

PR #2797, opened 2020-03-30
3.16

iwoodferren:


ws/src/main/webapp/WEB-INF/schemas/beans.xsd
```diff
@@ -38,6 +38,7 @@
         <sequence>
             <element name="label" type="beans:customStringType" minOccurs="0"/>
             <element name="secondaryLabel" type="beans:customStringType" minOccurs="0"/>
+            <element name="subjectOid" type="beans:customStringType"/>
```
svadla-oc: can you call this just "OID" to be consistent with naming convention of other fields?

svadla-oc:
👍

PR #2798, opened 2020-03-30
OC-12577 Participant Matrix: Event Status and the corresponding Icons are reversed

jkeremian:
Participant Matrix: Event Status and the corresponding Icons are reversed


svadla-oc:
👍

PR #2804, opened 2020-04-01
fix OC-12447 Verify button does not work on SDV Item Details table

rooney:
https://jira.openclinica.com/browse/OC-12447


svadla-oc:
👍

PR #2808, opened 2020-04-01
fix OC-12493: SDV Detailed View: Close option not visible if the page…

rooney:
… is minimized

https://jira.openclinica.com/browse/OC-12493


svadla-oc:
👍

PR #2810, opened 2020-04-01
OC-12593 Unable to enter data if the Participant is signed

jkeremian:
Unable to enter data if the Participant is signed


svadla-oc:
👍

PR #2818, opened 2020-04-06
OC-12396 reverting changes for triggering action every minute

jkeremian:


src/main/java/core/org/akaza/openclinica/web/helper/EventData.java
```diff
@@ -0,0 +1,50 @@
+package core.org.akaza.openclinica.web.helper;
```
svadla-oc, line 1: did you commit these files by mistake?

svadla-oc:
I am merging this since we want to run nightly against these changes and these new files don't break anything

PR #2828, opened 2020-04-08
OC-12504 Form version selected for data entry, always opens the default version of the form

jkeremian:
Form version selected for data entry, always opens the default version of the form


svadla-oc:
👍

PR #2830, opened 2020-04-09
OC-12596 Item data table for sdv should display item label for single…

rooney:
… and multi select for value column

https://jira.openclinica.com/browse/OC-12596


svadla-oc:
👍

PR #2831, opened 2020-04-10
OC-12440 Update SDV Item Details table header layout: fix Event start…

rooney:
… date's time

https://jira.openclinica.com/browse/OC-12440

### DIFF
--- a/src/main/java/org/akaza/openclinica/controller/dto/SdvDTO.java
+++ b/src/main/java/org/akaza/openclinica/controller/dto/SdvDTO.java
@@ -7,7 +7,7 @@ public class SdvDTO {
     private String participantId;
     private String siteName;
     private String eventName;
-    private Date eventStartDate;
+    private String eventStartDate;
     private Boolean eventStartDateHasTime;
     private int eventOrdinal;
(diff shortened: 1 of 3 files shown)


svadla-oc:
👍

src/main/java/org/akaza/openclinica/controller/dto/SdvDTO.java
```diff
@@ -7,7 +7,7 @@
     private String participantId;
     private String siteName;
     private String eventName;
-    private Date eventStartDate;
+    private String eventStartDate;
```
svadla-oc, line 10: How come changing this to a string fixes the problem? Ideally, the dto should keep the date type and the javascript code should handle the conversion properly.

rooney: hi Shiva, the Date arrives on the browser as int. Momentjs when given int will do timezone conversion, but won't if given String, so that fixed it. But OK, I'll revert the .java changes and handle all on client side.

rooney:
new PR: https://github.com/OpenClinica/OpenClinica/pull/2839

PR #2835, opened 2020-04-10
OC-10309 When user removes and restores a form under a signed event then event should become unsigned

jkeremian:
When user removes and restores a form under a signed event then event should become unsigned


svadla-oc:
👍

PR #2841, opened 2020-04-15
OC-12433 SDV Item Data Table not showing any item data or record meta…

rooney:
…data

https://jira.openclinica.com/browse/OC-12433

### DIFF
--- a/src/main/java/org/akaza/openclinica/controller/SdvApiController.java
+++ b/src/main/java/org/akaza/openclinica/controller/SdvApiController.java
@@ -19,6 +19,9 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
+
 import javax.servlet.http.HttpServletRequest;
 
 @Controller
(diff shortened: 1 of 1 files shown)


src/main/java/org/akaza/openclinica/controller/SdvApiController.java
```diff
@@ -47,29 +50,37 @@
...
+            }
+            return new ResponseEntity<>(responseDTO, HttpStatus.OK);    
         }
-        catch(OpenClinicaSystemException e) {
+        catch (Exception e) {
+            StringWriter errors = new StringWriter();
+            e.printStackTrace(new PrintWriter(errors));
             e.printStackTrace();
-            return new ResponseEntity<>(validateService.getResponseForException(e, studyOID, ""), HttpStatus.BAD_REQUEST);
+            return new ResponseEntity<>(errors.toString(), HttpStatus.BAD_REQUEST);
```
svadla-oc, line 82: I don't think it's a good idea to return the stacktrace in the API response. If the API is already returning 400, my guess is this is failing even before this method is executed. Have you checked the server logs to see if there are any errors there?

rooney: Hi Shiva, I don't have access to the server log. How to check it? Let's discuss on Slack.

PR #2844, opened 2020-04-16
fix OC-12596 Item data table for sdv should display item label for si…

rooney:
…ngle and multi select for value column

https://jira.openclinica.com/browse/OC-12596


svadla-oc:
👍

PR #2846, opened 2020-04-17
fix OC-12435: Opening form from SDV table and then closing it takes t…

rooney:
…he user to the Queries table

https://jira.openclinica.com/browse/OC-12435


svadla-oc:
👍

PR #2850, opened 2020-04-22
OC-12442:fix the new found broken issues in pipe delimited and pdf casebook ca…

Taoli2018:
…used by status refactor


svadla-oc:
👍

PR #2858, opened 2020-04-28
OC-12668 Opening an Event/Form popup on the Participant Matrix should close any popups that are already open

theywa:
[OC-12668](https://jira.openclinica.com/browse/OC-12668)


svadla-oc:
👍

PR #2861, opened 2020-04-28
OC-12641 Form status appearing incorrect in case book for completed forms

jkeremian:
Form status appearing incorrect in case book for completed forms


svadla-oc:
👍

PR #2870, opened 2020-05-01
OC1231 Solution fix

iwoodferren:
The resolution looks to use the EventCRFBean instead of the ItemDataBean when setting the request attributes. This allows for specific targeting of the event as well as properly retrieves the Study Subject's ID from the DB.


svadla-oc:
👍

PR #2883, opened 2020-05-05
OC-12744 Complete

iwoodferren:


web/src/main/java/org/akaza/openclinica/control/submit/DataEntryServlet.java
```diff
@@ -5753,9 +5753,11 @@ public void checkUpdateDataPermission(HttpServletRequest request) throws Insuffi
             }
 
             // continue to check the left open tab, it  maybe in different study
+            //This value comes from the event_definition_crf table to handle site-specific crfs
             String studyId = (String)request.getParameter("sid");
-            //The study id is always the study level id of the study, never the site.
-            if(studyId !=null && Integer.parseInt(studyId) != currentStudyParentId) {
+            //This checks both study and site ids against the event_definition_crf table's study_id value
+            //since crf's can be assigned to a specific site - zique
+            if(studyId !=null && Integer.parseInt(studyId) != currentStudyParentId && Integer.parseInt(studyId) != currentStudy.getId()) {
```
svadla-oc, line 5760: Consider this scenario 
1. User A is at study level for study X and had two tabs open - tab 1 with a form opened and tab 2 with some other page.
2. User A logs out of tab 2 
3. User B logs in to tab 2
4. User B switches to a site within the same study
5. User B goes to tab 1 and enters data for study X

With the logic you have above, the if condition would evaluate to false and allow User B to enter data into the form. I think you need to determine if "studyId" is a study or a site id and then check that value against the corresponding current study/site ids.

svadla-oc:
👍

PR #2885, opened 2020-05-05
OC-11023:Subject Matrix defaults showing only 15 Subjects

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
```diff
@@ -94,9 +94,34 @@
...
+        session = request.getSession();      
+     
+        // new setting
+        TableFacade tableFacade = null;
+        String matchedStudyOID = null;
+        if(request.getParameter("maxRows") != null) {        	
+        	tableFacade = prepareNewTableFacade(request, response);
+        }else {
+        	// restore from the session   	        	 
+        	 TableFacade tableFacadeInSession = (TableFacade) request.getSession().getAttribute("LSStableFacade");
```
svadla-oc, line 106: create a static variable to store the attribute name since it is referenced in multiple places. 
E.g. private static final String LIST_SUBJECTS_TABLE_FACADE = "LSStableFacade";

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
```diff
@@ -94,9 +94,34 @@
...
+        	 tableFacade = getTableFacadeImpl(request, response);
+        	 matchedStudyOID = (String) request.getSession().getAttribute("matchedStudyOID");
```
svadla-oc, line 108: same comment as above

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
```diff
@@ -94,9 +94,34 @@
...
+        	 String currentStudyOID = getStudyBean().getOid();
+        	 
+        	 if(matchedStudyOID != null && matchedStudyOID.equals(currentStudyOID)) {
+        		 if(tableFacadeInSession != null) {
+            		 Limit limitInSession = tableFacadeInSession.getLimit();
+        	         
+            		 tableFacade.setStateAttr("restore");
```
svadla-oc, line 115: does this need to have a name more specific to the matrix so it doesn't interfere with other tables?

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
```diff
@@ -94,9 +94,34 @@
...
+            		 tableFacade.setLimit(limitInSession);
```
svadla-oc, line 116: why not replace the entire table facade instead of just reusing the limit?

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
Taoli2018: done

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
Taoli2018: done

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
Taoli2018: tableFacade.setStateAttr("restore")  --- this is existing code/logic, and "restore" is under specfic tableFacade.

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectTableFactory.java
Taoli2018: I tried this ideal, but not working, will throw out all kinds of errors, I will sent your separate documents about those errors

svadla-oc: From what I see in the document you sent, it looks like you are trying to set items/state attributes on the retrieved table facade. What I am suggesting is to reuse the tableFacade as is in line 129.

Taoli2018: 1.I also tried that way -- the send document didn't include all solution I tried, what I found is that in  tableFacade it include lots of information like items, limits, maxrows.... but the maxrows per page is dynamically calculated from limits, so  when reuse the  tableFacade, the maxrows is always 15, even I reset as 50 like  tableFacade.setMaxRows(50) still not working, it still displat 15 on the UI; 2. the other big issue is that when reuse the tableFacade, if you change study, if the 2 study has different subjects count --- you will find that the list all subjects page may mess up, why because tableFacade  include lots of information like items, limits, maxrows, total rows... so if first study has 2000 subjects, when switch to 2nd study, which has only 100 subjects----  but will display as total 2000 subjects, so for sure we can't resue the whole tableFacade , we have to just partially use, like use the maxrows in the limit

svadla-oc: did you try appending studyId to the state attribute? something like "studyX-restore" so that the facade attributes are specific to a study? It seems to me that we are not using the library properly. 
I am ok merging the code you have currently but it would be great if we could persist all table settings.

Taoli2018: after check OC4, it has another solution, so to be consistent, I use  the solution in OC4

svadla-oc:
👍

PR #2903, opened 2020-05-11
OC-12715 Remove/Restore button is missing for common event for signed participant

jkeremian:
Remove/Restore button is missing for common event for signed participant


svadla-oc:
👍

PR #2906, opened 2020-05-12
Oc 12587:Item with Data Type of "Real" Accepts Letters

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/form/Validator.java
```diff
@@ -1214,6 +1214,10 @@ protected boolean isNumber(String fieldName) {
         }
         try {
             float f = Float.parseFloat(fieldValue);
+           
+            if(fieldValue.toLowerCase().endsWith("d") || fieldValue.toLowerCase().endsWith("f") || fieldValue.toLowerCase().endsWith("l") || fieldValue.toLowerCase().indexOf("e") > 0) {
+        		return false;
+        	}
```
svadla-oc, line 1220: can you use this method instead of the custom logic? https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/math/NumberUtils.html#isCreatable-java.lang.String-

I think this whole method can be replaced by the apache commons method

Taoli2018: I checked the document and also tested at my local, that method didn't meet our requirements
===========================================================
public static boolean isCreatable(String str)
Checks whether the String a valid Java number.

Valid numbers include hexadecimal marked with the 0x or 0X qualifier, octal numbers, scientific notation and numbers marked with a type qualifier (e.g. 123L).
======================================================
so any valid Java number, will pass that method test, like 123L, 309f.....
but the ticket complain this, they don't want to any accepts Letter in the real number

svadla-oc: Got it.

svadla-oc:
👍

PR #2907, opened 2020-05-12
OC-12437 Handle View button from SDV Item table when item is not visi…

rooney:
…ble on the form

https://jira.openclinica.com/browse/OC-12437


svadla-oc:
👍

PR #2914, opened 2020-05-13
OC-12785 Oops error is reported when changing event statuses through UI

jkeremian:
Oops error is reported when changing event statuses through UI


svadla-oc:
👍

PR #2915, opened 2020-05-13
OC-12674: Boolean var (actionsAllowed) is added in SdvItemDto for item with calculate type

suthirr-openclinica:


src/main/java/org/akaza/openclinica/controller/dto/SdvItemDTO.java
```diff
@@ -19,6 +19,15 @@
     private String lastModifiedUserLastName;
     private String sdvStatus;
     private int openQueriesCount;
+    private boolean isActionsAllowed;
```
svadla-oc, line 22: Instead of calling it actionsAllowed, I would just call it "isCalculateItem" and then have the UI handle it properly. I don't think we want to mask the reason for why the actions allowed from the UI.

suthirr-openclinica: Sure Shiva, I will change that

PR #2916, opened 2020-05-13
OC-12674: Boolean var (isCalculatedItem) is added in SdvItemDto for item with calculate type

suthirr-openclinica:


svadla-oc:
👍

PR #2933, opened 2020-05-18
OC-12540: Removed bug for not showing sdvItemTable when the study or site is with status other than Available

suthirr-openclinica:


svadla-oc:
👍

PR #2936, opened 2020-05-18
OC-12789: Removing unused line thats causing null error

slclinica:


src/main/java/core/org/akaza/openclinica/dao/extract/OdmExtractDAO.java
```diff
@@ -3431,7 +3431,6 @@ protected String studyEventAndFormMetaTables() {
     }
 
     protected String studyEventAndFormMetaCondition(int parentStudyId, int studyId, boolean isIncludedSite, String[] showArchived, String permissionTags) {
-        permissionTags = (permissionTags.length() != 0 ? permissionTags : "null");
```
svadla-oc, line 3434: should the method argument be removed as well?

PR #2939, opened 2020-05-19
OC-12777 Complete

iwoodferren:


web/src/main/java/org/akaza/openclinica/control/submit/ViewRuleAssignmentTableFactory.java
```diff
@@ -968,6 +971,12 @@ public void setDesignerLink(String designerLink)
         this.designerLink = designerLink;
     }
 
+    public String encodeTarget(String target){
+        target= StringUtils.replace(target,"[", "%5B");
```
svadla-oc, line 975: You should use this method so it can handle other characters as well https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html#encode-java.lang.String-java.lang.String-

svadla-oc:
👍

PR #2940, opened 2020-05-19
OC-12778 Complete

iwoodferren:


web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableClassic.java
```diff
@@ -1811,7 +1811,7 @@ public String stripQuotes(String subj) {
         } else {
             for (int i = 0; i < subjarray.length - 1; i++) {
                 returnme += subjarray[i];
-                returnme += "''";
```
svadla-oc, line 1814: looks like this code hadn't changed in a long time, so how is this not working now?

web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableRepeating.java
```diff
@@ -2799,7 +2799,7 @@ public String stripQuotes(String subj) {
         } else {
             for (int i = 0; i < subjarray.length - 1; i++) {
                 returnme += subjarray[i];
-                returnme += "''";
```
svadla-oc, line 2802: same comment as above

web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableClassic.java
iwoodferren: Yea this code/method I don't think is needed necessarily because of this commit:
https://github.com/OpenClinica/OpenClinica/commit/9310494488fe684dc7df82d8fd92ce38bb5923b7.
By switching over to using SqlParameter objects and using the PreparedStatement set methods, strings with SQL syntax characters are being escaped automatically before executing a DB update. The old code is there to escape those SQL syntax characters when directly executing a String as a DB query.

svadla-oc:
👍

PR #2941, opened 2020-05-19
OC-12495: crf status is checked for locked flag in studyEvent and the display name is set common value for sdvItem table and sdvData table

suthirr-openclinica:


svadla-oc:
👍

PR #2943, opened 2020-05-19
OC-12780:Can Change Repeating Event to Non-Repeating Even if There ar…

Taoli2018:
…e Already Multiple Occurrences


svadla-oc:
👍

PR #2944, opened 2020-05-20
OC-12605 value should wrap in the SDV data table, so the user should …

rooney:
…not have to scroll

https://jira.openclinica.com/browse/OC-12605


svadla-oc:
👍

PR #2953, opened 2020-05-21
OC-12778 Strip quotes method

iwoodferren:
Removed the stripQuotes method out of SpreadSheetTableClassic and 'Repeating.


web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableRepeating.java
```diff
@@ -2810,7 +2810,7 @@ public String stripQuotes(String subj) {
             returnme += subjarray[subjarray.length - 1];
         }
         return returnme;
-    }
+    }*/
```
svadla-oc, line 2813: It's better to delete the code that is not being used instead of commenting it out. We can always look at git history to figure out why this method was removed.

iwoodferren: Ok it's all set.

svadla-oc:
👍

PR #2954, opened 2020-05-21
OC-12828, OC-12830, OC-12765 and OC-12857

slclinica:
OC-12828: Change job detail identifier to job_uuid and change API to use job uuid instead of job name. 
OC-12830: Add job uuid to single jobs page. 
OC-12765: Fix oops error when editing job name. 
OC-12857: Show validation errors on update job page

### DIFF
--- a/src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
+++ b/src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
@@ -50,10 +50,10 @@ public class ScheduledExtractController {
 
 
     @ApiOperation(value = "To get latest scheduled extract dataset ids and creation time for the job name at study level", notes = "only work for authorized users with the right access permission")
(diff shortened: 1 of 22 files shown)


src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
```diff
@@ -50,10 +50,10 @@
 
 
     @ApiOperation(value = "To get latest scheduled extract dataset ids and creation time for the job name at study level", notes = "only work for authorized users with the right access permission")
-    @RequestMapping(value = "/studies/{studyOID}/extractJobs/{jobName}/jobExecutions", method = RequestMethod.GET)
+    @RequestMapping(value = "/studies/{studyOID}/extractJobs/{jobUuid}/jobExecutions", method = RequestMethod.GET)
```
svadla-oc, line 53: Since the jobUuid is unique, I don't think we need to force the user to provide studyOID as well. I would update the url to "extractJobs/{jobUuid}/jobExecutions"

svadla-oc:
👍

PR #2955, opened 2020-05-21
OC-12495 & OC-12-12497: CrfStatus changed to LockStatus and eventStatus is changed and filtered correctly

suthirr-openclinica:


svadla-oc:
👍

PR #2956, opened 2020-05-25
Issue 126

GerbenRienk:
When in view-study-subject you can collapse the icon-info, but you can't expand it afterwards. The reason for this is the double definition of the javascript-function leftnavExpand in the jsp, which is different from the one in global_functions_javascript.js


GerbenRienk:
Sorry, my mistake

PR #2964, opened 2020-05-28
OC-12873 signing event or participant is causing an oops error

theywa:
[OC-12873](https://jira.openclinica.com/browse/OC-12873)


svadla-oc:
Addressed with https://github.com/OpenClinica/OpenClinica/pull/2966

PR #2965, opened 2020-05-28
OC 12827 Sign action should not be shown for visit events when forms in the event is still in data entry started

theywa:
[#OC-12827](https://jira.openclinica.com/browse/OC-12827)


svadla-oc:
👍

PR #2967, opened 2020-05-28
OC-12755, OC-12804, OC-12756, OC-12801

slclinica:
OC-12755: Preserve past schedules, update table in extract tables to show action buttons only if file reference is present. 
OC-12804: Update table in single view job to show from archived_dataset_file db. 
OC-12756: Added manual vs schedule column. Made actions column always inline.
OC-12801: Add field to configure number of file references to keep in update/create job extract page.


svadla-oc:
👍

PR #2968, opened 2020-05-28
OC-12716 & OC-12747: Changed the default sdv value as null and changed the auditpage for accessing the correct sdv values

suthirr-openclinica:
OC-12716: Sdv status is changed to null when the event is created and then changed to corresponding sdv status after first sdv verifcation
OC-12747: Altered the audit page to get the correct sdvStatus values.


svadla-oc:
👍

svadla-oc:
👍

PR #2975, opened 2020-05-29
OC-12855: EventStartDate is wrongly gets updated as Null value when EventStatus is locked - is fixed

suthirr-openclinica:


svadla-oc:
👍

PR #2979, opened 2020-06-01
OC-11408: Duplicated Audit log entries for participant creation are stopped and workaround made for existing entries

suthirr-openclinica:


svadla-oc:
👍

PR #2980, opened 2020-06-02
OC-12437  Handle View button from SDV Item table when item is not visible on the form

rooney:
https://jira.openclinica.com/browse/OC-12437


svadla-oc:
👍

PR #2984, opened 2020-06-03
OC-12789

slclinica:
OC-12789: Fix null pointer by retrieving permission tags without request. Fix SPSS and SAS zip file generation.
OC-12796: Fix page breaking when search in ViewSingleJob page. (Partial fix, search still does not work)

### DIFF
--- a/src/main/resources/core/properties/archived_dataset_file_dao.xml
+++ b/src/main/resources/core/properties/archived_dataset_file_dao.xml
@@ -28,7 +28,7 @@
     <query>
         <name>findByJobUuidWithFileReference</name>
         <sql>SELECT * FROM ARCHIVED_DATASET_FILE
-            WHERE JOB_UUID=? AND FILE_REFERENCE IS NOT NULL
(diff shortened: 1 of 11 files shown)


src/main/resources/core/properties/archived_dataset_file_dao.xml
```diff
@@ -28,7 +28,7 @@
     <query>
         <name>findByJobUuidWithFileReference</name>
         <sql>SELECT * FROM ARCHIVED_DATASET_FILE
-            WHERE JOB_UUID=? AND FILE_REFERENCE IS NOT NULL
+            WHERE JOB_UUID=? AND FILE_REFERENCE != ''
```
svadla-oc, line 31: why is the file reference blank and not null? You should set the file reference to null instead of changing this.

slclinica: They are set to "". Same as how manual extracts set them when they are in queue.

svadla-oc:
👍

PR #2989, opened 2020-06-04
OC-12918 Removing Common event is not unsigning a signed common event

jkeremian:
Removing Common event is not unsigning a signed common event


svadla-oc:
👍

PR #2994, opened 2020-06-04
OC-12895 Fix and optimization

iwoodferren:
Old code implemented a hard stop at processing 15 rows of data to preserve speed. This was marked by variables firstLoopBreak and secondLoopBreak. How these breaks worked was they were skipping any new data somewhere around 60 whenever the user removed 15+ points of new data. Removing these loopBreaks fixed the issue of new data going missing, but introduced an exponential growth rate of time to save data. This proof is provided by the benchmarks below using a form similar to the customer's.

This PR provides a fix for both.

Benchmarks pre-fix:
100 new data points - 1+ minutes
100 existing data points + 20 new data points - 3 minutes

Benchmarks post-fix:
100 new data points - <= 1 minute
100 existing data points + 20 new data points - 1.5 minutes


svadla-oc:
👍

PR #3003, opened 2020-06-08
OC 12906 signing event brings signing participant page instead of signing events page in QE Dev

theywa:
[OC-12906](https://jira.openclinica.com/browse/OC-12906)

### DIFF
--- a/src/main/java/core/org/akaza/openclinica/service/DiscrepancyNoteUtil.java
+++ b/src/main/java/core/org/akaza/openclinica/service/DiscrepancyNoteUtil.java
@@ -4,13 +4,13 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 
+import javax.servlet.ServletContext;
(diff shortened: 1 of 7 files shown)


src/main/java/core/org/akaza/openclinica/service/DiscrepancyNoteUtil.java
```diff
@@ -128,6 +128,7 @@ public void injectParentDiscNotesIntoDisplayStudyEvents(List<DisplayStudyEventBe
         for (DisplayStudyEventBean dStudyEventBean : displayStudyBeans) {
             studyEventBean = dStudyEventBean.getStudyEvent();
             // All EventCRFs for a study event
+            eventCrfDAO = new EventCRFDAO(dataSource);
```
svadla-oc, line 131: we shouldn't be initializing DAOs like this anymore. Instead do something like below:
eventCRFDAO = (EventCRFDAO) SpringServletAccess.getApplicationContext(context).getBean("eventCRFJDBCDao");

See Jon's email for further information about this.

theywa: Even with the Qualifier annotation the eventCRFDAO still null, made some modification on local but it still null so I use the easiest way that you suggest bypassing the ServletContext to the method and use it to get the DAO.

svadla-oc:
👍

PR #3006, opened 2020-06-08
OC-12789: Fix scheduled extract errors

slclinica:
1. Added job name to the email error message
2. Removed delete archived dataset file functionality for scheduled jobs
3. Show failed when job fails instead of leave IN_PROGRESS
4. Use local HashMap variable to retrieve user roles instead of global variable thats causing confusion in variable type

### DIFF
--- a/src/main/java/core/org/akaza/openclinica/dao/login/UserAccountDAO.java
+++ b/src/main/java/core/org/akaza/openclinica/dao/login/UserAccountDAO.java
@@ -123,17 +123,19 @@ public class UserAccountDAO extends AuditableEntityDAO {
 
     }
 
-    public void setRoleTypesExpected() {
+    public HashMap setRoleTypesExpected() {
         this.unsetTypeExpected();
+        HashMap roleTypes = new HashMap();
         // assuming select star from study_user_role
-        this.setTypeExpected(1, TypeNames.STRING);
-        this.setTypeExpected(2, TypeNames.INT);
-        this.setTypeExpected(3, TypeNames.INT);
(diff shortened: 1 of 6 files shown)


src/main/java/core/org/akaza/openclinica/dao/login/UserAccountDAO.java
```diff
@@ -123,17 +123,19 @@ public void setPrivilegeTypesExpected() {
 
     }
 
-    public void setRoleTypesExpected() {
+    public HashMap setRoleTypesExpected() {
         this.unsetTypeExpected();
```
svadla-oc, line 127: you don't need this anymore.

src/main/java/core/org/akaza/openclinica/job/XsltTransformJob.java
```diff
@@ -570,17 +570,28 @@ protected void executeInternal(JobExecutionContext context) {
             postErrorMessage(e.getMessage(), context);
             logger.error("Error executing extract", e);
             exceptions = true;
+        } catch (NullPointerException e) {
```
svadla-oc, line 573: why are you catching NPE? It's usually not a good idea to catch RuntimeExceptions

src/main/java/core/org/akaza/openclinica/dao/login/UserAccountDAO.java
slclinica: It's removed?

svadla-oc:
👍

PR #3007, opened 2020-06-08
OC-12044: org.apache.common.lang doesn't support non-ascii characters is fixed by using common.lang3

suthirr-openclinica:


svadla-oc:
👍

PR #3009, opened 2020-06-09
refactor method filterRuleSetsByGroupOrdinal(...), so the sql  may ju…

Taoli2018:
…st run once, instead of  thousands or millions

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/service/rule/RuleSetService.java
+++ b/core/src/main/java/org/akaza/openclinica/service/rule/RuleSetService.java
@@ -849,17 +849,22 @@ public class RuleSetService implements RuleSetServiceInterface {
      * @see org.akaza.openclinica.service.rule.RuleSetServiceInterface#filterRuleSetsByGroupOrdinal(java.util.List)
      */
     public List<RuleSetBean> filterRuleSetsByGroupOrdinal(List<RuleSetBean> ruleSets) {
-
-        for (RuleSetBean ruleSetBean : ruleSets) {
+  	         
+    	for (RuleSetBean ruleSetBean : ruleSets) {
             List<ExpressionBean> expressionsWithCorrectGroupOrdinal = new ArrayList<ExpressionBean>();
             
-            String itemOid = getExpressionService().getItemOid(ruleSetBean.getExpressions().get(0).getValue());
-            String itemGroupOid = getExpressionService().getItemGroupOid(ruleSetBean.getExpressions().get(0).getValue());
+           /**
+	       	 * get the itemDataSize by studyEventId,itemGroupOid,itemOid for the whole study
+	       	 * only run once
+	       	 */  
+	       	String itemOid = getExpressionService().getItemOid(ruleSetBean.getExpressions().get(0).getValue());
+	        String itemGroupOid = getExpressionService().getItemGroupOid(ruleSetBean.getExpressions().get(0).getValue());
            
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/akaza/openclinica/service/rule/RuleSetService.java
```diff
@@ -852,15 +852,36 @@ private void logExpressions(List<RuleSetBean> validRuleSets) {
...
                 String groupOrdinal = getExpressionService().getGroupOrdninalCurated(expression.getValue());
-                List<ItemDataBean> itemDatas = getItemDataDao().findByStudyEventAndOids(Integer.valueOf(studyEventId), itemOid, itemGroupOid);
+              
                 logger.debug("studyEventId {} , itemOid {} , itemGroupOid {} , groupOrdinal {} , itemDatas {}", new Object[] { studyEventId, itemOid,
                     itemGroupOid, groupOrdinal, itemDatas.size() });
 
+                String key = ruleSetBean.getStudyId() + studyEventId + itemGroupOid +itemOid;
+                
+                if(itemDatasHm.containsKey(key)) {
+                	itemDatas.add((ItemDataBean) itemDatasHm.get(key));
```
svadla-oc, line 876: Looks like we are using the "itemDatas" variable only to get the count of item data. Can we just write a count query instead of loading all item data into a hashmap?

Taoli2018: Done - write a count query instead of loading all item data into a hashmap

core/src/main/java/org/akaza/openclinica/service/rule/RuleSetService.java
```diff
@@ -849,29 +849,56 @@ private void logExpressions(List<RuleSetBean> validRuleSets) {
...
+                itemOid = getExpressionService().getItemOid(expression.getValue());
+                itemGroupOid = getExpressionService().getItemGroupOid(expression.getValue());
                 String groupOrdinal = getExpressionService().getGroupOrdninalCurated(expression.getValue());
-                List<ItemDataBean> itemDatas = getItemDataDao().findByStudyEventAndOids(Integer.valueOf(studyEventId), itemOid, itemGroupOid);
-                logger.debug("studyEventId {} , itemOid {} , itemGroupOid {} , groupOrdinal {} , itemDatas {}", new Object[] { studyEventId, itemOid,
-                    itemGroupOid, groupOrdinal, itemDatas.size() });
-
+              
+               
+                String key = studyEventId + itemGroupOid +itemOid;
```
svadla-oc, line 876: instead of grouping the results and getting a count by this key, can you pass this key into the query itself? That way, we wouldn't have to get multiple counts.

Taoli2018: if pass this key into the query itself, then this means also need to put the query in the same loop level of where we get the key: studyEventId + itemGroupOid +itemOid, this is in loop:  for (ExpressionBean expression : ruleSetBean.getExpressions()) {...}, 
 so this refactor  will only just change the query to get the item data count, instead of item data objects, but the major performance issue is still there, the similar get count sql query will run 100K times in this attached test case.

svadla-oc: as discussed, calculate the item count per study event by moving the logic from the first for loop into the for loop below.

Taoli2018: done, please review

svadla-oc:
👍

PR #3013, opened 2020-06-09
OC-12955: StringUtils.EscapeXml changed to suppport emoji characters in ODM auditLog item values

suthirr-openclinica:


svadla-oc:
👍

PR #3015, opened 2020-06-09
Kafka related fixes: SessionFactory was creating too many connections…

jmcinerney:
…, fixed a NPE in KafkaService


svadla-oc:
👍

PR #3031, opened 2020-06-12
OC-12906 signing event brings signing participant page instead of signing events page in QE Dev

theywa:
[OC-12906](https://jira.openclinica.com/browse/OC-12906)


svadla-oc:
👍

PR #3042, opened 2020-06-15
OC-12960: Refactor code that starts job scheduler when tomcat restarts

slclinica:


src/main/java/core/org/akaza/openclinica/service/SchedulerUtilServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+                    logger.error("Bean for scheduler is not able to accessed: ", e);
+
+                }
+            }
+        }
+        return jobScheduler;
+    }
+
+
+    public void createSchedulerFactoryBean(ApplicationContext applicationContext, String schema) {
```
svadla-oc, line 59: this method should use what is being used in ScheduledJobConfig and have ScheduledJobConfig call this method

svadla-oc:
👍

PR #3045, opened 2020-06-16
OC-12967:Reverted the servlet and servlet-mapping which got deleted as part of kafka code cleanup

suthirr-openclinica:


svadla-oc:
👍

PR #3046, opened 2020-06-16
OC-12867 When visit based event is stopped, its getting extra attribute locked="No" in participant casebook and audit log

jkeremian:
When visit based event is stopped, its getting extra attribute locked="No" in participant casebook and audit log


svadla-oc:
👍

PR #3056, opened 2020-06-18
OC-12421 query table is not filtering item values for integer, decimal, and multi-selects

jkeremian:
query table is not filtering item values for integer, decimal, and multi-selects


svadla-oc:
👍

PR #3058, opened 2020-06-18
OC-12985: Resolved the encoding issue in sdv page

suthirr-openclinica:


src/main/java/core/org/akaza/openclinica/dao/EventCRFSDVFilter.java
```diff
@@ -56,11 +56,18 @@ private String buildCriteria(String criteria, String property, Object value) {
...
+                            sdvStatusFilterArray.add(SdvStatus.NOT_VERIFIED.toString());
+                        else
+                            sdvStatusFilterArray.add(SdvStatus.getByI18nDescription(sdvStatus.trim()).toString());
                     }
                 }
-                else
-                    sdvStatusFilterArray.add(SdvStatus.getByI18nDescription(sdvStatusString).toString());
+                else{
+                    if(sdvStatusString == null || sdvStatusString.toLowerCase().contains("null"))
+                        sdvStatusFilterArray.add(SdvStatus.NOT_VERIFIED.toString());
```
svadla-oc, line 67: In the if condition above, you are already accessing "sdvStatusString" so if you need to add a null check for this, you should add it to the if condition above

src/main/java/core/org/akaza/openclinica/dao/EventCRFSDVFilter.java
```diff
@@ -49,18 +56,22 @@ public String execute(String criteria) {
     }
 
     private String buildCriteria(String criteria, String property, Object value) {
-        value = StringEscapeUtils.escapeSql(value.toString());
-        if (value != null) {
+        if(StringUtils.isNotBlank(value.toString())){
```
svadla-oc, line 59: if value is null, this would throw an NPE. You should first check if value is null before checking for blank string.

suthirr-openclinica: Made changes as per the comment

svadla-oc:
👍

PR #3066, opened 2020-06-22
OC-12979: Reverted to correct user permission for changeCrfVersion icon

suthirr-openclinica:
Changed as part of OC-12403


svadla-oc:
👍

PR #3072, opened 2020-06-24
OC-12659: Reverted OC-12563 fetching eventCrfs from EventCrfDao

suthirr-openclinica:
Since first Eventcrf is created after studyEvent is fetched from db and hence null when fetched studyEvent.getEventCrfs()


svadla-oc:
👍

PR #3073, opened 2020-06-24
OC-12799 Edit Action on forms needs to be removed when the corresponding event status is Stopped

theywa:
[OC-12799](https://jira.openclinica.com/browse/OC-12799)


svadla-oc:
👍

PR #3076, opened 2020-06-24
OC-12761 Signing Participants from Participant Matrix with refactored event statuses

jkeremian:
Signing Participants from Participant Matrix with refactored event statuses


svadla-oc:
👍

PR #3079, opened 2020-06-25
OC-9536 Runtime pages load with error about external font resource

theywa:
Remove redundant import that caused log error on console. Font source already handled on icomoon-style.css.
[OC-9536](https://jira.openclinica.com/browse/OC-9536)


svadla-oc:
👍

PR #3081, opened 2020-06-25
Revert "OC-12799 Edit Action on forms needs to be removed when the corresponding event status is Stopped"

svadla-oc:
Reverts OpenClinica/OpenClinica#3073


svadla-oc:
#3073 is breaking participant details page.

PR #3087, opened 2020-06-29
OC-12542: disabled save button until the page is loaded and added key in item for removing the issue#2

suthirr-openclinica:


svadla-oc:
👍

PR #3101, opened 2020-07-07
OC-11957 Download rule file links should not be available if there ar…

rooney:
…e NO rules to download

https://jira.openclinica.com/browse/OC-11957


svadla-oc:
👍

PR #3123, opened 2020-07-14
OC-11101: participant status is displayed as "Contact Info Entered" when entered through reserve form

suthirr-openclinica:


svadla-oc:
👍

PR #3124, opened 2020-07-14
OC-13088: Added allUserRoles session attribute inside changeStudyServlet

suthirr-openclinica:


src/main/java/org/akaza/openclinica/control/core/SecureController.java
```diff
@@ -1420,12 +1420,8 @@ protected CustomRole checkMatchingUuid(CustomRole customRole, ChangeStudyDTO cha
 
     protected void populateCustomUserRoles(CustomRole customRole, String username) {
         List<ChangeStudyDTO> byUser = getStudyDao().findByUser(username);
-        List<StudyEnvironmentRoleDTO> userRoles = (List<StudyEnvironmentRoleDTO>) session.getAttribute("allUserRoles");
-        if (userRoles == null) {
-            logger.error("******************userRoles should not be null");
-            ResponseEntity<List<StudyEnvironmentRoleDTO>> responseEntity = getStudyBuildService().getUserRoles(request, true);
-            userRoles = responseEntity.getBody();
-        }
+        ResponseEntity<List<StudyEnvironmentRoleDTO>> responseEntity = getStudyBuildService().getUserRoles(request, true);
```
svadla-oc, line 1423: I think this is going to fetch user roles from user-service on every page in Runtime. Since that is an expensive call, we should only do it when changing studies or on login.

suthirr-openclinica: Only the function is currently in SecureController, But it is currently called only for ChangeStudyServlet. Also I made sure to call it once only when the ChangeStudyServlet is called the first time and not during 'submit'. Shall I move it ChangeStudyServlet to avoid confusion.

svadla-oc: yes please

suthirr-openclinica: Merge was not done properly, I will close this ticket and make a new PR Shiva

suthirr-openclinica: I moved populateCustomUserRoles() to ChangeStudyServlet

svadla-oc:
👍

PR #3126, opened 2020-07-15
OC-12802 Add navigation to SBS from Runtime: fix for Site level

rooney:
Fix for 404 when on Site level:

https://gist.githubusercontent.com/jmcinerney/11969626606659e2fd9179d1da85e159/raw/32aca64b176d4f1d3723ba27242427b470ca2814/gistfile1.txt

### DIFF
--- a/src/main/java/org/akaza/openclinica/control/core/SecureController.java
+++ b/src/main/java/org/akaza/openclinica/control/core/SecureController.java
@@ -1542,7 +1542,7 @@ public abstract class SecureController extends HttpServlet implements SingleThre
         return (PermissionService) SpringServletAccess.getApplicationContext(context).getBean("permissionService");
     }
 
-    protected void setStudy(Study study, HttpSession session) {
+    protected void getCurrentBoardUrl(Study study, HttpSession session) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/akaza/openclinica/control/core/SecureController.java
```diff
@@ -1549,6 +1550,11 @@ protected void setStudy(Study study, HttpSession session) {
                 try {
                     boardUrl = getStudyBuildService().getCurrentBoardUrl(accessToken, study);
                     study.setBoardUrl(boardUrl);
+                    try {
+                        getStudyDao().update(study);
+                    }
+                    catch (StaleStateException e) {
```
svadla-oc, line 1556: we need to understand why this exception is thrown. It is not a good idea to catch this exception and ignore.

rooney: @svadla-oc I've refactored the updating part to separate method that can be called independently from setting session's current study. This fixes the StaleStateException, so I removed the catch-and-ignore block on this PR's latest commit.

svadla-oc:
👍

PR #3132, opened 2020-07-17
OC-13156 Fix border style issues in latest Chrome

theywa:
[OC-13156](https://jira.openclinica.com/browse/OC-13156)

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/include/navBar.jsp
+++ b/src/main/webapp/WEB-INF/jsp/include/navBar.jsp
@@ -147,6 +147,15 @@
         realInterval * 1000
     );
 
+	// hacking OC-13156 Fix border style issues in latest Chrome
+    jQuery( document ).ready(function() {
+        if (navigator.userAgent.toLowerCase().indexOf('chrome') > -1) {
+            jQuery('input[type="text"]').each(function(){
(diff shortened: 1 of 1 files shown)


src/main/webapp/WEB-INF/jsp/include/navBar.jsp
```diff
@@ -147,6 +147,15 @@
...
 
+	// hacking OC-13156 Fix border style issues in latest Chrome
+    jQuery( document ).ready(function() {
+        if (navigator.userAgent.toLowerCase().indexOf('chrome') > -1) {
+            jQuery('input[type="text"]').each(function(){
+                jQuery(this).css('background-color', 'revert');
+            });
+        }
+    });
+
```
svadla-oc, line 158: Instead of updating the css using javascript like this, can you update the css directly as mentioned here: https://support.google.com/chrome/thread/48974735?hl=en&msgid=49104946?

theywa: Tried but it does not remove the border and made specific CSS style for chrome also not working that's why I put it on the script for workaround solution since a lot critic happens maybe this feature will be disabled in the future.

svadla-oc:
👍

PR #3133, opened 2020-07-17
OC-12982 Complete

iwoodferren:
Modifications based on security scans

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/control/admin/CreateUserAccountServlet.java
+++ b/web/src/main/java/org/akaza/openclinica/control/admin/CreateUserAccountServlet.java
@@ -280,8 +280,8 @@ public class CreateUserAccountServlet extends SecureController {
                                 addPageMessage(respage.getString("there_was_an_error_sending_account_creating_mail"));
                             }
                         } else {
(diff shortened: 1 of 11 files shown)


web/src/main/resources/org/akaza/openclinica/i18n/words.properties
```diff
@@ -579,7 +579,7 @@ confirmation_screen = Confirmation Screen
 
 congratulations_your_spreadsheet = <b>Congratulations!</b> Your spreadsheet generated no errors.
 
-congratulations_your_spreadsheet_no_errors = <b>Congratulations!</b>  Your spreadsheet generated no errors.<br/><b>Please review it.<br> You can save it to database by clicking "Continue".</b>
```
svadla-oc, line 582: why are you removing these html tags? If these are causing security issues, we need to find another approach to preserve the formatting of this text.

iwoodferren: Hey Shiva. Yes they are changes due to the security fixes for the new CSP policies. Malcolm looked it over and he thinks the changes look better than the original.

PR #3096, opened 2020-07-02
OC-13058: Resolved crfStatus to show the eventCrfWorkflowStatus in Queries page (UI, downloads & print page)

suthirr-openclinica:

PR #3135, opened 2020-07-17
OC-13058: Reverted previous OC-13058 code because Removed status is not an EventCrfWorkflowStatus and crfStatus is set from WorkflowStatus

suthirr-openclinica:
Reverted: https://github.com/OpenClinica/OpenClinica/pull/3096


svadla-oc:
👍

PR #3136, opened 2020-07-17
OC-12650     New Audit Log Event Types for Status Refactoring

jkeremian:
New Audit Log Event Types for Status Refactoring


src/main/resources/migration/tenant/4.0-2020-07-06-OC-12650.xml
```diff
@@ -4,35 +4,58 @@
                    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
 
 
-    <changeSet  author="jkeremian"  id="OC-12650-01">
+    <changeSet  author="jkeremian"  id="OC-12650-001">
```
svadla-oc, line 7: I don't think you can just change the existing liquibase script. It will break the deployment. You would have to create a new script that reverts the changes you had made earlier.

svadla-oc:
👍

PR #3145, opened 2020-07-21
OC-13163 Display Removed instead of Event Workflow Status on Participant Matrix Event Popup

theywa:
[OC-13163](https://jira.openclinica.com/browse/OC-13163)


svadla-oc:
👍

PR #3147, opened 2020-07-22
OC-12663 Migrate existing event records to new status model

jkeremian:
Migrate existing event records to new status model


svadla-oc:
👍

PR #3162, opened 2020-07-24
OC-13088: Removed the Null condition to get all the latest values from userRoles

suthirr-openclinica:
In my last PR, When moving the populateCustomUserRoles() from SecureController to ChangeStudyServlet (https://github.com/OpenClinica/OpenClinica/pull/3124), wrongly copied the old code with null condition. So removed it now


svadla-oc:
👍

PR #3178, opened 2020-07-29
OC-13172 Cleared/Archived/Removed form is preventing an event from being signed.

theywa:
[OC-13172](https://jira.openclinica.com/browse/OC-13172)


svadla-oc:
👍

PR #3182, opened 2020-07-29
OC-13226: Added the query condition to check for non-archived and non-removed form/event for sdvPage

suthirr-openclinica:
…


svadla-oc:
👍

PR #3187, opened 2020-07-30
OC-13188 Archiving/Unarchiving forms or events should unsign Events and Participants

jkeremian:
Archiving/Unarchiving forms or events should unsign Events and Participants


svadla-oc:
👍

PR #3188, opened 2020-07-30
OC-13252: If lastVerifiedDate is null and eventCrfStatus is Changed Since Verified, It takes the eventCrf status for all the items in the form

suthirr-openclinica:


svadla-oc:
👍

PR #3189, opened 2020-07-30
OC-13242 Participant Matrix single event view is only showing first Participant duplicated many times.

jkeremian:
Participant Matrix single event view is only showing first Participant duplicated many times.


svadla-oc:
👍

PR #3195, opened 2020-07-31
OC-11101: Reverted old OC-11101 and added new user status CONTACT_INFO_ENTERED is added for contact details entered through reserved form

suthirr-openclinica:


svadla-oc:
👍

PR #3197, opened 2020-07-31
OC-13114 Migrate existing form records to new status model

jkeremian:
Migrate existing form records to new status model


svadla-oc:
👍

PR #3202, opened 2020-08-03
OC-13202 Events that are Removed and also Locked should be able to be unlocked

theywa:
[OC-13202](https://jira.openclinica.com/browse/OC-13202)


svadla-oc:
👍

PR #3215, opened 2020-08-04
OC-13265 Signature attestation is not appearing with correct case in audit log download.

jkeremian:
Signature attestation is not appearing with correct case in audit log download.


svadla-oc:
👍

PR #3218, opened 2020-08-05
OC-13270: Logfile is created before calling before xml request

suthirr-openclinica:
- Logfile is created before calling before xml request

- Added unexpected_error_occured msg for all the unresolved errors

- Error message for ParticipantId header not matching the mappingFile, if available


svadla-oc:
👍

PR #3221, opened 2020-08-05
OC-13224 Migrate existing event, form, and form version definitions to new status model

jkeremian:
Migrate existing event, form, and form version definitions to new status model


svadla-oc:
👍

PR #3230, opened 2020-08-06
OC-13298 Audit log is not preserving old value and new value in some scenarios in ODM after migration

jkeremian:
OC-13298 Audit log is not preserving old value and new value in some scenarios in ODM after migration
OC-12520  Maintain old status after refactor
This PR includes fix for both above issues


svadla-oc:
👍

PR #3234, opened 2020-08-07
OC-13185 Logic fix for Signing Participants from Participant Matrix with refactored event statuses

theywa:
[OC-13185](https://jira.openclinica.com/browse/OC-13185)


svadla-oc:
👍

PR #3237, opened 2020-08-07
Update Study.java to remove nullable constraint

jmcinerney:
Remove "nullable = false" constraint as studyEnvUuid is not set for sites (which are stored as study objects).


jmcinerney:
jenkins retest this please

PR #3238, opened 2020-08-07
Update Study.java nullable constraint on studyEnvUuid

jmcinerney:
Remove "nullable = false" constraint as studyEnvUuid is not set for sites (which are stored as study objects).


svadla-oc:
👍

PR #3239, opened 2020-08-07
OC-13260: As per Paul's suggestion, Removed the full page width, To have a dynamic table size depending on the content

suthirr-openclinica:


svadla-oc:
👍

PR #3248, opened 2020-08-11
fix OC-13313: Clear CRF from Event page does not display event status…

rooney:
… or correct form status

https://jira.openclinica.com/browse/OC-13313


svadla-oc:
👍

PR #3252, opened 2020-08-11
OC-13305:  Altered jsp to check Removed instead of removed & also added check for Removed studySubject status on other places

suthirr-openclinica:
Caused By OC-10160 Terms.properties


svadla-oc:
👍

PR #3260, opened 2020-08-12
OC-13312: Fix dataset exports by replacing study event query from sub…

slclinica:
…ject_event_status_id to workflow_status


svadla-oc:
👍

PR #3262, opened 2020-08-12
OC-13324: Refactored the code to use KeycloakClientImpl instead of Au…

svadla-oc:
…thzClient to prevent intermittent keycloak errors


jmcinerney:
:+1:

PR #3272, opened 2020-08-14
OC-13327 & OC-13345: Refactored pipe-delimited to work with the refactored status & Resolved flat file not importing properly if the form is cleared

suthirr-openclinica:


svadla-oc:
👍

PR #3277, opened 2020-08-14
OC-13319: Added condition to allow only xml import with initial data entry and completed status in form

suthirr-openclinica:


svadla-oc:
👍

PR #3283, opened 2020-08-17
OC-13352 Sign button missing on common events after participant is signed

jkeremian:
Sign button missing on common events after participant is signed


svadla-oc:
👍

PR #3285, opened 2020-08-17
OC-13354: Restricted enketo to Read only mode if participant is not Available and not signed

suthirr-openclinica:


svadla-oc:
👍

PR #3287, opened 2020-08-18
OC-13331 common events that are removed are maintaining their workflow status, though they have attribute removed=yes set

rooney:
https://jira.openclinica.com/browse/OC-13331


svadla-oc:
👍

PR #3289, opened 2020-08-18
OC-13347: Non-printable character from Unicode sometimes gets added, Which causes header not matching

suthirr-openclinica:


svadla-oc:
👍

PR #3297, opened 2020-08-19
OC-13378 Openclinica:EventName is missing in Full ODM extract after study migration to Stack 12

jkeremian:
Openclinica:EventName is missing in Full ODM extract after study migration to Stack 12


svadla-oc:
👍

PR #3301, opened 2020-08-19
OC-13379: Sdv display values in AuditLog for null is altered to show empty string

suthirr-openclinica:


suthirr-openclinica:
jenkins retest this please

PR #3315, opened 2020-08-24
OC-12136: StudyEvent for non-repeating common event is inserted only during the first itemData insertion

suthirr-openclinica:


svadla-oc:
👍

PR #3321, opened 2020-08-25
fix OC-13410: Not Started forms missing locked status icon when event…

rooney:
… is locked on Sign Participant page

https://jira.openclinica.com/browse/OC-13410


svadla-oc:
👍

PR #3322, opened 2020-08-25
OC-13406 & OC-13405: change event listener making two backend call causes showing values of previous popup studySubject related values as well as new values

suthirr-openclinica:


svadla-oc:
👍

PR #3329, opened 2020-08-27
fix OC-13417: Event can be marked Complete when Required form is Removed

rooney:
https://jira.openclinica.com/browse/OC-13417


svadla-oc:
👍

PR #3332, opened 2020-08-28
fix OC-13432: Update Removed icon hovertext to "Removed" on Sign Even…

rooney:
…t and Sign Participant pages

https://jira.openclinica.com/browse/OC-13432


svadla-oc:
👍

PR #3335, opened 2020-08-28
OC-13433: Added condition to not show removed button for locked event in participant matrix page and single event page

suthirr-openclinica:


svadla-oc:
👍

PR #3338, opened 2020-08-28
OC-13440: Check if admin user has access to study or site for api ext…

slclinica:
…racts

### DIFF
--- a/src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
+++ b/src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
@@ -1,11 +1,13 @@
 package org.akaza.openclinica.controller;
 
 import core.org.akaza.openclinica.bean.extract.ArchivedDatasetFileBean;
+import core.org.akaza.openclinica.bean.login.StudyUserRoleBean;
 import core.org.akaza.openclinica.bean.login.UserAccountBean;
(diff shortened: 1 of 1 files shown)


src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
```diff
@@ -133,4 +159,14 @@ private ScheduledExtractJobDetailDTO convertEntityToDTO(ArchivedDatasetFileBean
         return scheduledExtractJobDetailDTO;
     }
 
+    private boolean userHasAccessToStudyOrSite(List<StudyUserRoleBean> userRoles, String studyOid) {
+        Study studyLevel = studyDao.findPublicStudy(studyOid);
+        studyOid = studyOid.trim().toUpperCase();
+        if (studyLevel != null) {
+            return validateService.isUserHasAccessToStudy(userRoles, studyOid);
+        } else {
+            return validateService.isUserHasAccessToSite(userRoles, studyOid);
```
svadla-oc, line 168: studyOid will not be a site oid. Given the study oid, you need to check if user has access to any of the sites in that study.

svadla-oc: Also, if "studyLevel" is null, it could mean that the given study oid is invalid. We can't assume that it is a site.

PR #3341, opened 2020-08-31
OC-13448: Added condition to remove completed status option in UpdateStudyEvent page if the required form is not started

suthirr-openclinica:


svadla-oc:
👍

PR #3344, opened 2020-08-31
OC-13452 Should be able to complete event if required form is archived

jkeremian:
Should be able to complete event if required form is archived


svadla-oc:
👍

PR #3355, opened 2020-09-02
fix OC-13434: Unexpected "click more options" is showing in participa…

rooney:
…nt matrix event/form pop up windows

https://jira.openclinica.com/browse/OC-13434


svadla-oc:
👍

PR #3361, opened 2020-09-03
OC-13480: Added condition to check if site level required form eventDefinitionCrf is alreadyCreated

suthirr-openclinica:


svadla-oc:
👍

PR #3363, opened 2020-09-03
OC-13478 Monitors getting oops error when filtering single events in staging

jkeremian:
Monitors getting oops error when filtering single events in staging


svadla-oc:
👍

PR #3365, opened 2020-09-03
OC-13485: Added condition for updateStudyEvent to have completed status option if required form is not-started and archived

suthirr-openclinica:


svadla-oc:
👍

PR #3368, opened 2020-09-04
OC-13487: Used the same query for filter count and actual values for consistancy and Removed the duplicate query

suthirr-openclinica:


svadla-oc:
👍

PR #3383, opened 2020-09-09
OC-13514: Resolved the import with wrong formLayoutOID in common event failure with no error message

suthirr-openclinica:


svadla-oc:
@suthirr-openclinica this should be merged into 4.13.x branch.

PR #3385, opened 2020-09-10
OC-13250: Add synchronized to extract creation

slclinica:


src/main/java/core/org/akaza/openclinica/job/XsltTransformJob.java
```diff
@@ -116,7 +116,7 @@
     private static final long KILOBYTE = 1024;
 
     @Override
-    protected void executeInternal(JobExecutionContext context) {
```
svadla-oc, line 119: Making the whole method synchronized could have performance issues since only one thread can run this method at a given time. Can we narrow down the code that is causing this and make that code synchronized?

svadla-oc (changes requested):

svadla-oc:
👍

PR #3389, opened 2020-09-11
OC- 13273 Commas no longer separate selected multi select options in Audit log or Clear CRF From Event page

jkeremian:
Commas no longer separate selected multi select options in Audit log or Clear CRF From Event page


svadla-oc:
@jkeremian looks like this PR is created from the 4.11.x branch?

PR #3390, opened 2020-09-14
OC-13199 Update Status attribute to display on same line as workflow …

rooney:
…status on Participant Matrix

https://jira.openclinica.com/browse/OC-13199


svadla-oc:
👍

PR #3395, opened 2020-09-15
OC-12023: Updated last updated date and updated user in StudySubject in event and form changes

suthirr-openclinica:


svadla-oc:
Instead of handling this in multiple different places, can we check if data is changed in these entities at the Service level or Dao level and only then update the last modified date?

PR #3397, opened 2020-09-16
OC-13591 Matrix performance improvements

jmcinerney:


src/main/java/core/org/akaza/openclinica/dao/managestudy/EventDefinitionCRFDAO.java
```diff
@@ -716,6 +716,11 @@ public EventDefinitionCRFBean findByStudyEventIdAndCRFVersionId(Study study, int
 		return edc;
 	}
 
+	public EventDefinitionCRFBean findByStudyEventIdAndCRFVersionId(int studyEventId, int crfVersionId) {
```
svadla-oc, line 719: why did you add this method? This seems to be just calling that other method.

src/main/java/core/org/akaza/openclinica/service/managestudy/StudySubjectServiceImpl.java
```diff
@@ -532,6 +544,44 @@ public void populateUncompletedCRFsWithCRFAndVersions(DataSource ds, ArrayList u
         }
     }
 
+
+
+    public Boolean isSignable(StudyUserRoleBean currentRole, int studySubjectId) {
```
svadla-oc, line 549: I think the role check should happen in the calling method. This method should just check if the subject is signable not include the check for if a user can sign the subject.

svadla-oc (changes requested):

svadla-oc:
👍

PR #3399, opened 2020-09-16
fix OC-13559: Headers for Archived Common Event displayed for Partici…

rooney:
…pants who have no data in those events

https://jira.openclinica.com/browse/OC-13559


svadla-oc:
👍

PR #3400, opened 2020-09-16
OC-13546: Remove data entry person access to access scheduled extract…

slclinica:
…s from API


src/main/java/org/akaza/openclinica/controller/ScheduledExtractController.java
```diff
@@ -84,6 +85,14 @@
             return ResponseEntity.badRequest().headers(HeaderUtil.createFailureAlert(ENTITY_NAME, ErrorConstants.ERR_INVALID_STUDY_OID,
                     "No study found for StudyOId:" + studyOid + ".")).body(null);
         }
+
+        StudyUserRoleBean studyRoleBean = userAccountBean.getRoleByStudy(study);
```
svadla-oc, line 89: This code should go into ValidateService. It's generally not a good idea to put business logic into controller classes since the logic cannot be reused.

svadla-oc (changes requested):

svadla-oc:
👍

PR #3403, opened 2020-09-17
OC-13526: Importing Queries functionality is added

suthirr-openclinica:
WIP

### DIFF
--- a/src/main/java/core/org/akaza/openclinica/bean/odmbeans/DiscrepancyNoteBean.java
+++ b/src/main/java/core/org/akaza/openclinica/bean/odmbeans/DiscrepancyNoteBean.java
@@ -19,6 +19,7 @@ import java.util.List;
  *
  */
 public class DiscrepancyNoteBean extends ElementOIDBean {
+    private String displayId;
     private String status;
     private String noteType;
     private Date dateUpdated;
@@ -81,5 +82,13 @@ public class DiscrepancyNoteBean extends ElementOIDBean {
 
     public void setThreadUuid(String threadUuid) {
         this.threadUuid = threadUuid;
+    }
+
+    public String getDisplayId() {
+        return displayId;
+    }
+
+    public void setDisplayId(String displayId) {
+        this.displayId = displayId;
     }
 }
--- a/src/main/java/org/akaza/openclinica/controller/openrosa/QueryServiceImpl.java
+++ b/src/main/java/org/akaza/openclinica/controller/openrosa/QueryServiceImpl.java
@@ -15,6 +15,8 @@ import javax.mail.MessagingException;
 import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
 import javax.mail.internet.MimeMessage;
+
+import core.org.akaza.openclinica.exception.OpenClinicaSystemException;
 import org.akaza.openclinica.controller.openrosa.processor.QueryServiceHelperBean;
 import core.org.akaza.openclinica.core.EmailEngine;
 import core.org.akaza.openclinica.core.form.xform.QueriesBean;
@@ -36,6 +38,7 @@ import core.org.akaza.openclinica.service.crfdata.EnketoUrlService;
 import core.org.akaza.openclinica.web.SQLInitServlet;
 import core.org.akaza.openclinica.web.pform.OpenRosaService;
 import core.org.akaza.openclinica.web.pform.StudyAndSiteEnvUuid;
+import org.akaza.openclinica.web.restful.errors.ErrorConstants;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang3.RandomUtils;
 import org.codehaus.jackson.map.ObjectMapper;
@@ -166,7 +169,7 @@ public class QueryServiceImpl implements QueryService {
         }
     }
 
-    public DiscrepancyNote createQuery(QueryServiceHelperBean helperBean, QueryBean queryBean,boolean parentDn) throws Exception {
+    public DiscrepancyNote createQuery(QueryServiceHelperBean helperBean, QueryBean queryBean,boolean parentDn) {
         DiscrepancyNote dn = new DiscrepancyNote();
         dn.setStudy(helperBean.getContainer().getStudy());
         dn.setEntityType("itemData");
@@ -453,12 +456,15 @@ public class QueryServiceImpl implements QueryService {
     public String generateDisplayId(Boolean parentDn){
         DiscrepancyNote dn = null;
         String newDisplayId =null;
+        int counter = 0;
         do {
             String randomNo = StringUtils.leftPad(Integer.toString(RandomUtils.nextInt(0, 1000000000)), 9, '0');
             newDisplayId = parentDn ? "DN_"+randomNo : "CDN_"+randomNo;
             dn = discrepancyNoteDao.findByDisplayId(newDisplayId);
-        }while(dn != null);
-
+            counter++;
+        }while(dn != null && counter < 10);
+        if(counter >= 10)
+            throw new OpenClinicaSystemException("Failed", ErrorConstants.ERR_GENERATING_DISCREPANCY_NOTE_ID);
         return newDisplayId;
     }
 }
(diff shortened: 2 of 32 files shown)


src/main/java/core/org/akaza/openclinica/bean/odmbeans/DiscrepancyNoteBean.java
```diff
@@ -82,4 +83,12 @@ public String getThreadUuid() {
     public void setThreadUuid(String threadUuid) {
         this.threadUuid = threadUuid;
     }
+
+    public String getDisplayId() {
```
svadla-oc, line 87: not sure how these get displayed in the ODM, but we want to use the existing "ID" attribute and not add a new attribute called "DisplayID". Only the database will have this new column and the "ID" attribute value will be populated from this column.

src/main/java/org/akaza/openclinica/service/ImportValidationServiceImpl.java
```diff
@@ -0,0 +1,109 @@
...
+
+        for(ChildNoteBean childNoteBean : discrepancyNoteBean.getChildNotes()){
+            if(!checkDiscrepancyNoteStatusValid(childNoteBean.getStatus()))
+                throw new OpenClinicaSystemException(FAILED, ErrorConstants.ERR_DISCREPANCY_NOTE_STATUS_NOT_VALID);
+            if(!isUserExist(childNoteBean.getOwnerUserName()))
+                throw new OpenClinicaSystemException(FAILED, ErrorConstants.ERR_USER_NOT_VALID);
+            if(childNoteBean.getUserRef() != null && !isUserExist(childNoteBean.getUserRef().getUserName()))
+                throw new OpenClinicaSystemException(FAILED, ErrorConstants.ERR_ASSIGNED_USER_NOT_VALID);
+            if(StringUtils.isBlank(childNoteBean.getDetailedNote()))
+                throw new OpenClinicaSystemException(FAILED, ErrorConstants.ERR_DETAILED_NOTE_NOT_AVAILABLE);
```
svadla-oc, line 56: instead of throwing exception for each error, we should collect all all errors in a collection and throw the exception at the end with all errors.

src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -333,11 +340,19 @@ public boolean validateAndProcessDataImport(ODMContainer odmContainer, String st
...
-                                    continue;
                                 } else if (itemObject instanceof Item) {
                                     item = (Item) itemObject;
                                 }
-
+                                for(DiscrepancyNoteBean discrepancyNoteBean : itemDataBean.getDiscrepancyNotes().getDiscrepancyNotes()){
+                                    try {
+                                        importValidationService.validateQuery(discrepancyNoteBean);
+                                        createQuery(discrepancyNoteBean, tenantStudy, studySubject, eventCrf, itemDataBean.getItemOID(), Integer.parseInt(itemGroupDataBean.getItemGroupRepeatKey()));
+                                    }catch (Exception e){
```
svadla-oc, line 350: You should only catch the OpenClinicaSystemException since that's what you are expecting. All runtime exceptions should be just thrown and maybe caught at a higher level if needed.

src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -1432,47 +1447,95 @@ private Object validateItem(ImportItemDataBean itemDataBean, CrfBean crf, EventC
...
+                    lastComment = childDN.getDetailedNotes();
+
+                    // update Item data map
+                    helperBean.setDn(childDN);
+                    helperBean.setParentDn(parentDn);
+                    queryService.saveQueryItemDatamap(helperBean);
+                }
+                parentDn.setDetailedNotes(lastComment);
+                parentDn.setUserAccountByOwnerId(lastuser);
+                discrepancyNoteDao.saveOrUpdate(parentDn);
```
svadla-oc, line 1533: move the code to create a child note to a new method so it's easier to maintain and potentially you can reuse this code for creating parent note as well?

svadla-oc (changes requested):

src/main/java/org/akaza/openclinica/service/ImportValidationServiceImpl.java
```diff
@@ -0,0 +1,395 @@
...
+                return validateForDate(value);
+            default:
+                itemCountInForm.setInsertedUpdatedSkippedItemCountInForm(itemCountInForm.getInsertedUpdatedSkippedItemCountInForm() + 1);
+                throw new OpenClinicaSystemException(FAILED, ErrorConstants.ERR_ITEM_TYPE_NOT_SUPPORTED);
+        }
+    }
+
+    private Boolean validateForBoolean(String value) {
+        if (!value.equals("true") && !value.equals("false")) {
+            return false;
```
svadla-oc, line 364: Instead of returning false throw the errror

src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -1432,90 +1305,115 @@ private Object validateItem(ImportItemDataBean itemDataBean, CrfBean crf, EventC
...
+                helperBean.setDn(discNote);
+                discNote = discrepancyNoteDao.saveOrUpdate(discNote);
+                queryService.saveQueryItemDatamap(helperBean);
+                DataImportReport dataImportReport =  new DataImportReport(studySubject.getOcOid(), studySubject.getLabel(),
+                        eventCrf.getStudyEvent().getStudyEventDefinition().getOc_oid(), eventCrf.getStudyEvent().getSampleOrdinal().toString(),
+                        eventCrf.getFormLayout().getCrf().getOcOid(), itemGroupDataBean.getItemGroupOID(), itemGroupDataBean.getItemGroupRepeatKey(), itemOid, discNote.getDisplayId()+" "+INSERTED, sdf_logFile.format(new Date()),  null);
+                dataImportReports.add(dataImportReport);
+            }
+            if (isParentDn)
+            {
```
svadla-oc, line 1384: move the curly brace to the same line as if statement

src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
```diff
@@ -1432,90 +1305,115 @@ private Object validateItem(ImportItemDataBean itemDataBean, CrfBean crf, EventC
...
+                            DataImportReport dataImportReport =  new DataImportReport(studySubject.getOcOid(), studySubject.getLabel(),
+                                    eventCrf.getStudyEvent().getStudyEventDefinition().getOc_oid(), eventCrf.getStudyEvent().getSampleOrdinal().toString(),
+                                    eventCrf.getFormLayout().getCrf().getOcOid(), itemGroupDataBean.getItemGroupOID(), itemGroupDataBean.getItemGroupRepeatKey(), itemOid, discNote.getDisplayId()+" "+UPDATED, sdf_logFile.format(new Date()),  null);
+                            dataImportReports.add(dataImportReport);
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            logger.error("Query Import Failed: ", e);
```
svadla-oc, line 1403: Translate the errors from queryService to a more user-friendly error messages

svadla-oc (changes requested):

src/main/java/org/akaza/openclinica/service/ImportServiceImpl.java
suthirr-openclinica: I checked the query service code, The exception is not mandatory in those places, It was added to throw error if the container doesn't have the value required, I will remove the throws Exception, Since it will anyways get caught in the parent call.

src/main/java/org/akaza/openclinica/controller/openrosa/QueryServiceImpl.java
```diff
@@ -432,15 +438,27 @@ private void sendEmail(String to, String subject, String body, Boolean htmlEmail
...
     }
+
+    public String generateDisplayId(Boolean parentDn){
+        DiscrepancyNote dn = null;
+        String newDisplayId =null;
+        do {
+            String randomNo = StringUtils.leftPad(Integer.toString(RandomUtils.nextInt(0, 1000000000)), 9, '0');
+            newDisplayId = parentDn ? "DN_"+randomNo : "CDN_"+randomNo;
+            dn = discrepancyNoteDao.findByDisplayId(newDisplayId);
+        }while(dn != null);
```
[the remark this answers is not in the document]
suthirr-openclinica: I have added condition to check 10 times

src/main/java/core/org/akaza/openclinica/bean/odmbeans/DiscrepancyNoteBean.java
suthirr-openclinica: I made changes to display it as "ID" attribute

svadla-oc:
👍

PR #3408, opened 2020-09-22
OC-13553: Altered the updateId properly to fetch created user Id properly when updated user Id is not present

suthirr-openclinica:


jmcinerney:
:+1:

PR #3412, opened 2020-09-28
OC-13556: Updated id throwing null pointer exception, and failing to update studySubject is fixed

suthirr-openclinica:


svadla-oc:
👍

PR #3420, opened 2020-10-01
OC-13526: Altered discrepancy note error messages, fixed a bug on looking for existing discrepancy_note & added condition to continue if value is not present in itemData

suthirr-openclinica:


svadla-oc:
👍

PR #3421, opened 2020-10-01
OC-9830 In Dataset create and edit, not all CRF Versions are listed for items in more than on version

jkeremian:
In Dataset create and edit, not all CRF Versions are listed for items in more than on version


svadla-oc:
👍

PR #3422, opened 2020-10-02
OC-13613: Altereed query download to get Note id from displayId & removed the unused DownLoadBean interface 

suthirr-openclinica:


svadla-oc:
👍

PR #3439, opened 2020-12-07
when only 1 occurrence is scheduled and has data entered, still be able to be changed to non repeating

Taoli2018:


svadla-oc:
👍

PR #3440, opened 2020-12-16
fixed Not Started filter

Taoli2018:


Taoli2018:
please see https://github.com/OpenClinica/OpenClinica/pull/3468

PR #3443, opened 2020-12-17
OC-12780:when only 1 occurrence is scheduled and has data entered

Taoli2018:
it needs to still be able to be changed to non repeating

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/dao/managestudy/StudyEventDAO.java
+++ b/core/src/main/java/org/akaza/openclinica/dao/managestudy/StudyEventDAO.java
@@ -1277,6 +1277,7 @@ public class StudyEventDAO extends AuditableEntityDAO implements Listener {
 
         if (it.hasNext()) {
             Integer count = (Integer) ((HashMap) it.next()).get("count");
+            //count >0 means that we found at least one subject who has been scheduled this repeating event more than once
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/akaza/openclinica/dao/managestudy/StudyEventDAO.java
```diff
@@ -1270,7 +1277,7 @@ public boolean isThisRepeatingEventScheduledMoreThanOneTime(int studyId,int sed_
 
         if (it.hasNext()) {
             Integer count = (Integer) ((HashMap) it.next()).get("count");
-            if(count.intValue() >= 1) {
+            if(count.intValue() > 1) {
```
svadla-oc, line 1280: please refer to the comment in #3444 and reconcile the changes

svadla-oc (changes requested):

core/src/main/java/org/akaza/openclinica/dao/managestudy/StudyEventDAO.java
Taoli2018: Now change to ount.intValue() > 0, same logic as #3444.
because count >0 means that we found at least one subject who has been scheduled this repeating event more than once

svadla-oc:
👍

PR #3444, opened 2020-12-22
OC-12780: solve conflicts, more code change after testing

Taoli2018:


core/src/main/java/org/akaza/openclinica/dao/managestudy/StudyEventDAO.java
```diff
@@ -1277,7 +1277,7 @@ public boolean isThisRepeatingEventScheduledMoreThanOneTime(int studyId,int sed_
 
         if (it.hasNext()) {
             Integer count = (Integer) ((HashMap) it.next()).get("count");
-            if(count.intValue() > 1) {
+            if(count.intValue() > 0) {
```
svadla-oc, line 1280: I don't think we want to change this logic. Perhaps the calling method is not handling the result of this method properly

svadla-oc (changes requested):

core/src/main/java/org/akaza/openclinica/dao/managestudy/StudyEventDAO.java
Taoli2018: I double check the code, count >0 means that we found at least one subject who has been scheduled this repeating event more than once, count >1 means that we found at least two subjects who has been scheduled this repeating event more than once, so should change the logic here

svadla-oc: Sounds good.

svadla-oc:
👍

PR #3453, opened 2021-04-09
fixed OC-13705 and also fixed OC-14727

Taoli2018:


svadla-oc:
Please revert to 3.14 code and check how the emojis were breaking extracts and if it is possible to escape them without actually displaying the emojis.

Taoli2018:
I reverted to 3.14, and original implementation has escape all emojs like below:
<ItemData ItemOID="I_3999T_TEXT1" Value="&#55357;&#56837;&#55357;&#56845;&#55357;&#56836;"/>
and this cause the xml process failed, unicode has same format like &#nnnnn, current open lib for xml process can support unicode, but when the number is out of some scope like some emojs, it failed, I also added all detail errors and debug info which point to where failed in the code
in ticket :https://jira.openclinica.com/browse/OC-7263

Taoli2018:
please see https://github.com/OpenClinica/OpenClinica/pull/3464

PR #3454, opened 2021-04-20
OC-5701:fixed -- implement a new feature of  tab-separated or   pipe-separate…

Taoli2018:
…d file file

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/bean/extract/DownloadDiscrepancyNote.java
+++ b/core/src/main/java/org/akaza/openclinica/bean/extract/DownloadDiscrepancyNote.java
@@ -42,6 +42,8 @@ public class DownloadDiscrepancyNote implements DownLoadBean{
     public static String CSV ="text/plain; charset=UTF-8";
     public static String PDF = "application/pdf";
     public static String COMMA = ",";
+    public static String PIPE = "|";
+    public static String TAB = "\t";
     public static Map<Integer,String> RESOLUTION_STATUS_MAP = new HashMap<Integer,String> ();
     static{
         RESOLUTION_STATUS_MAP.put(1,"New");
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/akaza/openclinica/bean/extract/DownloadDiscrepancyNote.java
```diff
@@ -514,6 +677,62 @@ public void downLoadDiscBeans(List<DiscrepancyNoteBean> listOfBeans,
 
     }
 
+    public void downLoadThreadedDiscBeans(List<DiscrepancyNoteThread> listOfThreadedBeans,
```
svadla-oc, line 680: Instead of creating a new method like this, update the existing method to take delimiter as argument and the value for it would be null for PDF formats.

core/src/main/java/org/akaza/openclinica/bean/extract/DownloadDiscrepancyNote.java
```diff
@@ -188,6 +190,167 @@ public int getThreadListContentLength(List<DiscrepancyNoteThread> threadBeans) {
...
+
+
+        writer.append("\n");
+        return writer.toString();
+
+
+    }
+
 
     public String serializeToString(EntityBean bean, boolean includeHeaderRow,
```
svadla-oc, line 355: add the delimiter as argument here instead of the method above

svadla-oc (changes requested):

core/src/main/java/org/akaza/openclinica/bean/extract/DownloadDiscrepancyNote.java
Taoli2018: done -- please see changes in new PR:
https://github.com/OpenClinica/OpenClinica/pull/3463

core/src/main/java/org/akaza/openclinica/bean/extract/DownloadDiscrepancyNote.java
Taoli2018: done -- please see changes in new PR:
https://github.com/OpenClinica/OpenClinica/pull/3463

Taoli2018:
done -- please see changes in new PR:
https://github.com/OpenClinica/OpenClinica/pull/3463

so close this duplicated one

PR #3463, opened 2021-05-06
OC-5701:fix --- implemented a new feature of tab-separated or pipe-se…

Taoli2018:
…parated file file, so if user has issue when download the comma-separated file, then they can select different format to bypass the data conflicts.


Taoli2018:
this is code change based on code review feedback in the PR #3454

PR #3464, opened 2021-05-10
OC-13705:fixed unicode and emoji (emoji will only display as it'scode…

Taoli2018:
… as plain text)

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/bean/extract/odm/ClinicalDataReportBean.java
+++ b/core/src/main/java/org/akaza/openclinica/bean/extract/odm/ClinicalDataReportBean.java
@@ -186,11 +186,8 @@ public class ClinicalDataReportBean extends OdmXmlReportBean {
                             } 
                             if(printValue) {
                                 Boolean hasElm = false;
-                                String value = StringEscapeUtils.escapeXml(item.getValue());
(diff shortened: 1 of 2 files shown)


Taoli2018:
code change after code review  in #3452

Taoli2018:
new PR:
https://github.com/OpenClinica/OpenClinica/pull/3464

core/src/main/java/org/akaza/openclinica/bean/extract/odm/ClinicalDataReportBean.java
```diff
@@ -595,5 +584,25 @@ public int compare(Object o1, Object o2) {
 	    });
 	}
 	
+	/**
+	 *  get emoji as plain text code
+	 * @param s
+	 * @return
+	 */
+	private String processEmoji(String s) {
```
svadla-oc, line 592: instead of processing the emoji, we may be able to fix it by upgrading xercesImpl. OC4 doesn't seem to have this issue

Taoli2018: after check the code , in OC4 the solution is to use comons-lang3, now change the OC3ocde to use the same solution

svadla-oc:
👍

PR #3465, opened 2021-05-13
OC-12777:fixed the issue -- when there are square brackets in the rule…

Taoli2018:
… mssage,  can't open square rule deisgner


svadla-oc:
👍

PR #3469, opened 2021-05-24
OC-15058:fix the issue of no longer throwing a warning message when r…

Taoli2018:
…esponse options are changed between versions

### DIFF
--- a/web/src/main/java/org/akaza/openclinica/control/admin/CreateCRFVersionServlet.java
+++ b/web/src/main/java/org/akaza/openclinica/control/admin/CreateCRFVersionServlet.java
@@ -684,31 +684,26 @@ public class CreateCRFVersionServlet extends SecureController {
     public ResponseOptionBean hasDifferentOption(ResponseSetBean oldRes, ResponseSetBean newRes) {
         ArrayList oldOptions = oldRes.getOptions();
         ArrayList newOptions = newRes.getOptions();
-        if (oldOptions.size() != newOptions.size()) {
-            // if the sizes are different, means the options don't match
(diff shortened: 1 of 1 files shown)


web/src/main/java/org/akaza/openclinica/control/admin/CreateCRFVersionServlet.java
```diff
@@ -686,7 +686,41 @@ public ResponseOptionBean hasDifferentOption(ResponseSetBean oldRes, ResponseSet
...
         if (oldOptions.size() != newOptions.size()) {
             // if the sizes are different, means the options don't match
-            return null;
+        	//OC-15058, but still need to compare the "overlapped" part
+        	int smallerSize =0;
+        	if(oldOptions.size() < newOptions.size()) {
+        		smallerSize = oldOptions.size();
+        	}else {
+        		smallerSize = newOptions.size();
+        	}
```
svadla-oc, line 695: If the sizes are different, should we throw an error?

svadla-oc: What happens if the order of the values change?

Taoli2018: Get  more feedback about the requirements from CS team:
If the sizes are different, should we throw an error? ---- NO, still allow to process.
What happens if the order of the values change? --- Order doesn't matter as long as the coded values are kept the same
Did more code changes after reconsider above feedback.

svadla-oc:
👍

PR #3471, opened 2021-05-24
OC-9170:Marking a CRF Complete breaks navigating between sections of CRF

Taoli2018:
please see old PR #1447

### DIFF
--- a/web/src/main/webapp/includes/global_functions_javascript.js
+++ b/web/src/main/webapp/includes/global_functions_javascript.js
@@ -1,3 +1,22 @@
+function checkDataStatus() {
+
+        objImage=document.getElementById('status_top');
+        if (objImage != null && objImage.src.indexOf('images/icon_UnsavedData.gif')>0) {
+            return confirm('<fmt:message key="you_have_unsaved_data" bundle="${resword}"/>');
+        }
+
+        return true;
+    }
+    
+function gotoLink() {
+
+        var OptionIndex=document.crfForm.sectionName.selectedIndex;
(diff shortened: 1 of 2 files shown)


web/src/main/webapp/includes/global_functions_javascript.js
```diff
@@ -1,3 +1,22 @@
+function checkDataStatus() {
+
+        objImage=document.getElementById('status_top');
+        if (objImage != null && objImage.src.indexOf('images/icon_UnsavedData.gif')>0) {
+            return confirm('<fmt:message key="you_have_unsaved_data" bundle="${resword}"/>');
+        }
+
+        return true;
+    }
```
svadla-oc, line 9: It looks like viewSectionDataEntry.jsp had commented out checkDataStatus function. These changes will bring back the checkDataStatus function for that jsp. We should verify why that function was commented out and whether it's ok to bring it back.

svadla-oc: There are some jsps that have this javascript file imported and also have these functions defined within the jsp file. We should make sure that these functions don't interfere with each other. E.g. administrativeEditing.jsp has this scenario.

svadla-oc (changes requested):

Taoli2018:
we did more testing on current code, even after roll back all code changes of "fix", still can't reproduce this issue any more, so this may related to temporarily browser render issue, at this time, we decide to not commit any code changes, if it happened again in the future, we'll come back to revisit this issue.

PR #3472, opened 2021-05-24
OC-8255:Manual CRF import creates duplicate event CRF if a different CRF version already has been started for that event.

Taoli2018:
please see old PR #3069

PR #3152, opened 2020-07-22
OC-8239:WS CRF import creates duplicate event CRF if a different CRF …

Taoli2018:
…version already has been started for that event.

PR #3473, opened 2021-05-24
OC-8239:WS CRF import creates duplicate event CRF if a different CRF version already has been started for that event.

Taoli2018:
Please see #3152 
and this PR need to merge after #3472 merged, because they share some code


svadla-oc:
👍

PR #3157, opened 2020-07-23
OC-8254:Scheduled CRF import creates duplicate event CRF if a differe…

Taoli2018:
…nt CRF version already has been started for that event.

PR #3474, opened 2021-05-26
OC-8254:Scheduled CRF import creates duplicate event CRF if a different CRF version already has been started for that event

Taoli2018:


Taoli2018:
please also see #3157 for 3.16

svadla-oc:
👍

PR #3475, opened 2021-06-03
OC-14264:Upgrade OC 3.16 Postgres to 11

Taoli2018:
upgrade liquibase to latest version to fix reported issue

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/dao/core/OCContextLoaderListener.java
+++ b/core/src/main/java/org/akaza/openclinica/dao/core/OCContextLoaderListener.java
@@ -5,7 +5,7 @@ import java.net.UnknownHostException;
 
 import javax.servlet.ServletContextEvent;
 
-import liquibase.log.LogFactory;
+import liquibase.logging.LogFactory;
 
 import org.slf4j.MDC;
 import org.slf4j.bridge.SLF4JBridgeHandler;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/akaza/openclinica/dao/core/OCContextLoaderListener.java
```diff
@@ -34,7 +34,7 @@ public void contextInitialized(ServletContextEvent event) {
         // MDC.put("WEBAPP", webAppName + " FROM " + hostName);
         // Get the liquibase logs inside the application log files using
         // SLF4JBridgeHandler
-        LogFactory.getLogger().addHandler(new SLF4JBridgeHandler());
+        //LogFactory.getLogger().addHandler(new SLF4JBridgeHandler());
```
svadla-oc, line 37: Can you verify that the logging still works after commenting this out? We may have to enable logging a different way with the new version. See https://stackoverflow.com/questions/20880783/how-to-get-liquibase-to-log-using-slf4j

Taoli2018: I did testing after implement like https://stackoverflow.com/questions/20880783/how-to-get-liquibase-to-log-using-slf4j
I found the liquibase log behavior  has no any change ----- still only log info/message in the console when first time start tomcat to create database, and I also did testing in old version without any liquibase/postgres  upgrade, I found that the liquibase log behavior is also same, so thoise changes has no any impact on the liquibase log functionalities.

svadla-oc:
👍

PR #3481, opened 2021-07-29
OC-12780: more code change after testing, make sure that the new added…

Taoli2018:
… fix logic will only trigger when Change Repeating Event to Non-Repeating Event


svadla-oc:
👍

PR #3506, opened 2022-03-07
As Microsoft are currently deprecating TLS versions 1.0 and 1.1,so se…

Taoli2018:
…t default value  as 1.2

### DIFF
--- a/core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
+++ b/core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
@@ -417,13 +417,13 @@ public class CoreResources implements ResourceLoaderAware {
          * set default value
          */
         if(DATAINFO.getProperty("mailSmtpStarttls.required") == null) {
-        	DATAINFO.setProperty("mail.smtp.starttls.required", "true");
+        	DATAINFO.setProperty("mail.smtp.starttls.required", "");
         }else {
         	DATAINFO.setProperty("mail.smtp.starttls.required", DATAINFO.getProperty("mailSmtpStarttls.required"));
         }
         
         if(DATAINFO.getProperty("mailSmtpSslProtocols") == null) {
-        	DATAINFO.setProperty("mail.smtp.ssl.protocols", "TLSv1.2");
+        	DATAINFO.setProperty("mail.smtp.ssl.protocols", "");
         }else {
         	DATAINFO.setProperty("mail.smtp.ssl.protocols", DATAINFO.getProperty("mailSmtpSslProtocols"));
         }
--- a/core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
+++ b/core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
@@ -416,15 +416,11 @@ public class CoreResources implements ResourceLoaderAware {
          * As Microsoft are currently deprecating TLS versions 1.0 and 1.1
          * set default value
          */
-        if(DATAINFO.getProperty("mailSmtpStarttls.required") == null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
```diff
@@ -412,8 +412,22 @@ private void setMailProps() {
...
+        if(DATAINFO.getProperty("mailSmtpStarttls.required") == null) {
+        	DATAINFO.setProperty("mail.smtp.starttls.required", "true");
+        }else {
+        	DATAINFO.setProperty("mail.smtp.starttls.required", DATAINFO.getProperty("mailSmtpStarttls.required"));
+        }
+        
+        if(DATAINFO.getProperty("mailSmtpSslProtocols") == null) {
+        	DATAINFO.setProperty("mail.smtp.ssl.protocols", "TLSv1.2");
+        }else {
+        	DATAINFO.setProperty("mail.smtp.ssl.protocols", DATAINFO.getProperty("mailSmtpSslProtocols"));
```
svadla-oc, line 428: I don't know if all our customers would be able to upgrade to TLSv1.2. I think we should set these properties only if the customer explicitly sets them. Otherwise, we should skip setting these properties.

svadla-oc (changes requested):

core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
Taoli2018: changed -- we should set these properties only if the customer explicitly sets them. Otherwise, we should skip setting these properties.

core/src/main/java/org/akaza/openclinica/dao/core/CoreResources.java
```diff
@@ -417,13 +417,13 @@ private void setMailProps() {
...
         if(DATAINFO.getProperty("mailSmtpStarttls.required") == null) {
-        	DATAINFO.setProperty("mail.smtp.starttls.required", "true");
+        	DATAINFO.setProperty("mail.smtp.starttls.required", "");
         }else {
         	DATAINFO.setProperty("mail.smtp.starttls.required", DATAINFO.getProperty("mailSmtpStarttls.required"));
         }
         
         if(DATAINFO.getProperty("mailSmtpSslProtocols") == null) {
-        	DATAINFO.setProperty("mail.smtp.ssl.protocols", "TLSv1.2");
+        	DATAINFO.setProperty("mail.smtp.ssl.protocols", "");
```
svadla-oc, line 426: Setting them to blank might have an undesirable effect. I would just not set the properties at all. E.g. I would change the if-else block here to:
`if(DATAINFO.getProperty("mailSmtpSslProtocols") != null) {
DATAINFO.setProperty("mail.smtp.ssl.protocols", DATAINFO.getProperty("mailSmtpSslProtocols"));
}
`

svadla-oc: It looks like there are two ways of setting these properties. One via spring using applicationContext-email.xml and other in EmailEngine.props. The ones that are set here need to be read in EmailEngine class and added to EmailEngine.props. 
For emails sent using spring, you need to set the defaults within applicationContext-email.xml
E.g. `<prop key="mail.smtp.ssl.protocols">s[mail.smtp.ssl.protocols: TLSv1.1]</prop>`
Reference: https://stackoverflow.com/a/2534949

Taoli2018: code change like below: we should set these properties only if the customer explicitly sets them. Otherwise we will use default setting of these 2 properties in applicationContext-core-email.xml.

svadla-oc:
👍

PR #3509, opened 2022-03-18
Because branch 3.15.8 never push to production, so we don't bounce mi…

Taoli2018:
…nor version number after security patch


svadla-oc:
👍

PR #3516, opened 2022-04-15
OC-17705:Fix XSS finding in report

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/form/FormProcessor.java
```diff
@@ -132,7 +145,38 @@ public String getString(String fieldName, boolean searchAttributes) {
...
+        int scriptStart = result.indexOf("<script>");
+        int scriptEnd = result.lastIndexOf("</script>");
+        if(scriptStart > -1 && scriptEnd > -1 && scriptEnd >= scriptStart) {        
+        	escapedStr= result.substring(0, scriptStart) + result.substring(scriptEnd);
+        }
+        
+        escapedStr= escapedStr.replaceAll("<script>", "");
+        escapedStr= escapedStr.replaceAll("</script>", "");
+        escapedStr= escapedStr.replaceAll("<style>", "");
+        escapedStr= escapedStr.replaceAll("</style>", "");
```
svadla-oc, line 166: I am not sure we need to replace these tags if you are escaping javascript in the line below

svadla-oc: In general, this seems like a big change and I am wary of the consequences of this change. I think we should limit the fix to the affected code by escaping javascript in the jsp that is displaying the query parameters. E.g. using a c:out tag will automatically escape the values.

svadla-oc (changes requested):

web/src/main/java/org/akaza/openclinica/control/form/FormProcessor.java
Taoli2018: I agree that this'll be a big change if fix like this at code level, in this reported case, we use one 3rd party technology ---Jmesa dynamic tables ([jmesa-2.4.2-oc.jar](https://code.google.com/archive/p/jmesa/source)), so all page html code (include javascripts, css) are dynamically generated by the 3rd party library with the HTTP Request as one of the input, for example, in java class file ListStudySubjectsServlet.java, under the method createTable() will use below line to create table html:

There is no place to use  c:out tag to intercept the output of the above method,
I will do more research at Nginx content security policy setting, hope can find solution by modification of nginx configuration

PR #3522, opened 2022-04-28
Revert " .OC-13471:Duplicate Subjects Created"

svadla-oc:
Reverts OpenClinica/OpenClinica#3379


svadla-oc:
Reverting since we are not including this fix in 3.14

PR #3533, opened 2022-06-23
OC-18299:Web service data import is failing for import that is workin…

Taoli2018:
…g on manual import

### DIFF
--- a/ws/src/main/java/org/akaza/openclinica/ws/DataEndpoint.java
+++ b/ws/src/main/java/org/akaza/openclinica/ws/DataEndpoint.java
@@ -224,6 +224,8 @@ public class DataEndpoint {
 
         String xml = node2String(odmElement);
         xml = xml.replaceAll("<ODM>", this.ODM_HEADER_NAMESPACE);
+        xml = xml.replaceAll("<!\\[CDATA\\[","");
+        xml = xml.replaceAll("\\]\\]>","");
 
         if (xml == null)
             throw new Exception(respage.getString("unreadable_file"));


ws/src/main/java/org/akaza/openclinica/ws/DataEndpoint.java
```diff
@@ -224,6 +224,8 @@ private ODMContainer unmarshallToODMContainer(Element odmElement) throws Excepti
 
         String xml = node2String(odmElement);
         xml = xml.replaceAll("<ODM>", this.ODM_HEADER_NAMESPACE);
+        xml = xml.replaceAll("<!\\[CDATA\\[","");
```
svadla-oc, line 227: CDATA is used to escape special characters in xml. If we remove it completely, this could break the xml parser. I would update the xml parser to handle CDATA. 
Please test with data that has special characters like "<", ">"

svadla-oc (changes requested):

ws/src/main/java/org/akaza/openclinica/ws/DataEndpoint.java
Taoli2018: After did more testing and check the code,

looks like the posted test case xml file and the related OC3 document doesn't match the current OC3 code,

if just simple remove the <!CDATA[* and *]]> tag in the original xml file which was posted in this ticket, should work, in this case no need to do any code change/fix.so just close the PR without merge, we may also need to update the related OC3 document later.

PR #3546, opened 2022-08-25
OC-12863:Inconsistency in upversioning forms with single quotation

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableClassic.java
```diff
@@ -973,6 +973,11 @@ public NewCRFBean toNewCRF(javax.sql.DataSource ds, ResourceBundle resPageMsg) t
                                 ownerId = oldItem.getOwner().getId();
                             }
                         }
+                        
+                        if(ownerId == 0) {
+                        	ownerId =  ub.getId();
+                        }
```
svadla-oc, line 979: if ownerId is 0, it could mean that ub was empty. Perhaps, another thread is resetting ub to empty object.

svadla-oc (changes requested):

web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableClassic.java
Taoli2018: When add a new from version with new item field, this will happen, here is detail process in current implementation:

ownerId is overriden on line 973:
ownerId = oldItem.getOwner().getId();

and trace the code, this is caused by line 892:
ItemBean oldItem = (ItemBean) idao.findByNameAndCRFId(itemName, crfId);

when use the itemName and crfId to search item from database, if can't find, then retem a blank item with ownerId=0

web/src/main/java/org/akaza/openclinica/control/admin/SpreadSheetTableClassic.java
```diff
@@ -973,6 +973,11 @@ public NewCRFBean toNewCRF(javax.sql.DataSource ds, ResourceBundle resPageMsg) t
                                 ownerId = oldItem.getOwner().getId();
```
svadla-oc, line 973: Add an if condition to only set ownerId to oldItem's owner id if it is not 0.

svadla-oc (changes requested):

svadla-oc (approved):

PR #3547, opened 2022-08-30
OC-18491:AE/SAE rule fires for old entries

Taoli2018:


core/src/main/java/org/akaza/openclinica/dao/submit/ItemDataDAO.java
```diff
@@ -181,11 +181,13 @@ public HashMap findByStudySubjectAndOids(Integer studyId, String itemOid, String
...
         while (it.hasNext()) {
         	HashMap hm = (HashMap) it.next();
-        	//construct key
-        	String key = (String) this.getKeyFromHashMap(hm);
         	
         	//construct dataItem
         	ItemDataBean eb = (ItemDataBean) this.getEntityFromHashMap(hm);	
+        	
+        	//construct full item data key, like:1435IG_ADVER_AE_2965[2].I_ADVER_AESAE
+      	    String key = (Integer) hm.get("study_event_id") + (String) hm.get("ig_ocoid")+"[" + eb.getOrdinal() +"]" + (String) hm.get("item_ocoid");
```
svadla-oc, line 189: instead of apply this change here, please update the logic in this.getKeyFromHashMap(hm) method.
Also, verify that this change will resolve issues with all calls that are made to this.getKeyFromHashMap(hm) method.

svadla-oc (changes requested):

core/src/main/java/org/akaza/openclinica/dao/submit/ItemDataDAO.java
Taoli2018: Hi Shiva,
 Could you please review same PR for 3.17.x https://github.com/OpenClinica/OpenClinica/pull/3544
I have made code changes based on you review feedback on 3.16.x
"instead of apply this change here, please update the logic in this.getKeyFromHashMap(hm) method.
Also, verify that this change will resolve issues with all calls that are made to this.getKeyFromHashMap(hm) method."

and I'll close 3.16.x PR: https://github.com/OpenClinica/OpenClinica/pull/3547
Because as you mentioned than we have no plan to do any new build/deployment on any older versions.

Thanks

Taoli2018:
close 3.16.x PR: https://github.com/OpenClinica/OpenClinica/pull/3547
Because as Shiva mentioned that we have no plan to do any new build/deployment on any older versions.
all fix will focus on 3.17.x

PR #3552, opened 2023-01-19
OC-17767:Check XSS finding in report

Taoli2018:


web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectsServlet.java
```diff
@@ -139,6 +140,19 @@ private void createTable() {
         factory.setStudyParameterValueDAO(getStudyParameterValueDAO());
         String findSubjectsHtml = factory.createTable(request, response).render();
 
+        /**
+         * check unsafe input
+         */
+        int index = findSubjectsHtml.indexOf("jQuery.jmesa.addFilterToLimit('findSubjects','studySubject.label',");
```
svadla-oc, line 146: Instead of fixing it specifically for the matrix, let's try to fix it globally like we did for OC4.

svadla-oc (changes requested):

web/src/main/java/org/akaza/openclinica/control/submit/ListStudySubjectsServlet.java
Taoli2018: I merged OC4 fix, and tested at my local, looks like the fix of OC4 can't fix the issue reported in OC3, OC4 fix is to handle XSS when user/attacker  enter value at input fields from matrix table, but OC3 reported the XSS issue is user/attacker start XSS attack from URL.

PR #3595, opened 2023-06-02
Downgrading spring library versions

kkrumlian:


core/src/main/java/org/akaza/openclinica/service/crfdata/DynamicsMetadataService.java
```diff
@@ -521,7 +520,13 @@ private String getValue(PropertyBean property, RuleSetBean ruleSet, EventCRFBean
         if(property.getValueExpression() == null) {
             logger.info("There is no ValueExpression for property ="+property.getOid());
              if (stratificationFactorBeans!=null)
-                 value=getRandomizeService().getRandomizationCode(eventCrfBean,stratificationFactorBeans,ruleSet);
+                try {
+                    value=getRandomizeService().getRandomizationCode(eventCrfBean,stratificationFactorBeans,ruleSet);
+                } catch (JSONException e) {
+                    // TODO Auto-generated catch block
+                    e.printStackTrace();
```
svadla-oc: Can we use logger to print this instead as this would get printed to tomcat log?

svadla-oc (changes requested):

core/src/main/java/org/akaza/openclinica/service/crfdata/DynamicsMetadataService.java
kkrumlian: This is fixed @svadla-oc !

svadla-oc (approved):

PR #3597, opened 2023-06-06
OC-21016 Allow partial dates searching

rehmanali-oc:


kkrumlian (commented):
What's the impact of this change ... how many places does it impact?

kkrumlian:
@rehmanali-oc ^^

rehmanali-oc:
> What's the impact of this change ... how many places does it impact?

This change has an impact on multiple classes/pages where this class is inherited. I will explain the issue that I am addressing with this change using some examples.

Let's consider the calculated search string for the view event page is `PID event_date event_status`. We split this search string by spaces into multiple substrings, and then we compare each substring with our search keyword. The problem arises when our search keyword contains a dash ("-"). For instance, if the search string is "07-May-2023", it would return false if our search keyword also contained a dash.

Here are some examples to illustrate the issue:
1. "07" returns true
2. "May" returns true
3. "2023" returns true
4. "07-May" returns false, along with other similar scenarios. This is because our logic splits the string by dashes and compares each substring individually, which leads to no results being returned if our keyword contains characters around the dashes as well.

By addressing this problem, the change aims to resolve the inconsistency in matching search keywords containing dashes and improve the accuracy of the search functionality.

kkrumlian (approved):

PR #3627, opened 2023-07-11
OC-21319 - Remove crf version error

qasimSM:
- NPE was being thrown when uploading a CRF with the same version name as the already uploaded CRF but with some items deleted (or missing).


qasimSM:
Wrong branch

PR #3628, opened 2023-07-11
OC-21319 - Fixed remove crf version error

qasimSM:
- NPE was being thrown when uploading a CRF with the same version name as the already uploaded CRF but with some items deleted (or missing)


qasimSM:
Closing as discussed in standup. A new dev bug is created to track this issue. See the comments on https://jira.openclinica.com/browse/OC-21319 for details.

PR #3638, opened 2023-07-18
OC-18491 Fix Rules firing for incorrect entries

rehmanali-oc:


core/src/main/java/org/akaza/openclinica/service/rule/expression/ExpressionService.java
```diff
@@ -460,6 +461,7 @@ public String evaluateExpression(String expression) throws OpenClinicaSystemExce
                 String valueFromDb = null;
                 String matchEvents = null;
                 String valueFromForm = null;
+                expression = map.get("expression");
```
kkrumlian: Rename this Map key to expressionWithFixedOrdinals.

kkrumlian (approved):

PR #3641, opened 2023-07-20
OC-21377 - LDAP user is not populated in create user form

rehmanali-oc:


web/src/main/java/org/akaza/openclinica/controller/user/LdapUserController.java
```diff
@@ -56,8 +65,13 @@ public String listLdapUsers(HttpServletRequest req,
     @RequestMapping("/selectLdapUser")
     public String selectLdapUser(HttpServletRequest req, @RequestParam(value="dn", required = false) String dn) {
         // A request without "dn" parameter is used to just close the iFrame
-        if (!StringUtils.isEmpty(dn)) {
-            req.getSession().setAttribute("ldapUser", ldapUserService.loadUser(dn));
```
kkrumlian, line 60: In order to ensure backwards compatibility let's continue to support dn and if not present use the new model of looking up users ...

kkrumlian (approved):

PR #3682, opened 2024-07-11
Fixing file downloads for windows/linux paths.

kkrumlian:


web/src/main/java/org/akaza/openclinica/control/submit/DownloadAttachedFileServlet.java
```diff
@@ -172,4 +180,16 @@ public void processRequest() throws Exception {
         }
     }
 
+    private boolean startsWithIgnoringSlashes(String str1, String str2) {
+        // Replace all forward slashes with empty strings
+        String normalizedStr1 = str1.replace("/", "").replace("\\", "");
```
svadla-oc, line 185: Instead of all these string manipulations in jsp and java code which could be prone to other issues, I think it would have been better to just encode the path in jsps and decode them in java. Can we do that?
Looking at Qasim's comment in the ticket, tomcat configuration update was another option due to security concerns. Since we are validating the path using canonical path that security risk is mitigated, should we just update the tomcat configuration instead?
Apparently browsers were supposed to encode these characters per the 3986 spec but since not all browsers are following that spec, tomcat added an option to allow backslash from v 9.0.8 onwards. See https://bz.apache.org/bugzilla/show_bug.cgi?id=62273

kkrumlian: This particular code is doing the following
- Ensuring that path that we are downloading the file from is the path of openclinica.data/attached_files path 
- This is in place to ensure someone doesn't change the path in URL and try to download another file from the system. 
- I would suggest we put this fix in, we are basically not touching the current fix that's in place. From the testing perspective the scope will be limited
- Even if we do the above we might need the above fix.

